PR #46, opened 2015-02-06
Python pylint pep8

benlast:
Python coding standards

Status: **Opened for review**

Reviewers: **Any and all Python devs**
JIRA: **https://mobify.atlassian.net/browse/CRTM-82**

## Changes
- Added a big README to cover Python standards and code checking
- Added a standardized `pylintrc`

## How to test-drive this PR
- Read the README
- Try `pep8`, `pylint` and `pyflakes`

### DIFF
--- a/python/pylintrc
+++ b/python/pylintrc
@@ -172,7 +172,7 @@ no-docstring-rgx=__.*__
 
 # Minimum line length for functions/classes that require docstrings, shorter
 # ones are exempt.
-docstring-min-length=-1
+docstring-min-length=-3
 
 
 [FORMAT]
--- a/python/pylintrc
+++ b/python/pylintrc
@@ -1,0 +1,371 @@
+[MASTER]
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Profiled execution.
+profile=no
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS,.git,node_modules,venv,puppet,.vagrant,build
+
+# Pickle collected data for later comparisons.
+persistent=no
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Use multiple processes to speed up Pylint.
+jobs=4
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
+
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED
+confidence=
+
+# Enable the message, report, category or checker with the given id(s). You can
+# either give multiple identifier separated by comma (,) or put this option
+# multiple time. See also the "--disable" option for examples.
+#enable=
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once).You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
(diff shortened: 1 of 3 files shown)


python/pylintrc
```diff
@@ -0,0 +1,371 @@
...
+[BASIC]
+
+# Required attributes for module, separated by a comma
+required-attributes=
+
+# List of builtins function names that should not be used, separated by a comma
+bad-functions=map,filter,input
+
+# Good variable names which should always be accepted, separated by a comma
+good-names=i,j,k,ex,Run,_
```
kyoung: Run?

python/pylintrc
```diff
@@ -0,0 +1,371 @@
...
+msg-template='{path}:line {line}: {obj}: {msg_id}:{symbol}:{msg}'
+
+
+[BASIC]
+
+# Required attributes for module, separated by a comma
+required-attributes=
+
+# List of builtins function names that should not be used, separated by a comma
+bad-functions=map,filter,input
```
kyoung: fp is considered harmful?

python/pylintrc
```diff
@@ -0,0 +1,371 @@
...
+# Naming hint for method names
+method-name-hint=[a-z_][a-z0-9_]{2,30}$
+
+# Regular expression which should only match function or class names that do
+# not require a docstring.
+no-docstring-rgx=__.*__
+
+# Minimum line length for functions/classes that require docstrings, shorter
+# ones are exempt.
+docstring-min-length=-1
```
kyoung: Will this require docstrings for any function longer than one line?

That might be a bit much.

python/pylintrc
```diff
@@ -0,0 +1,371 @@
...
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once).You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
+disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1626,W1607,W1606,W1605,W1604,W1603,W1602,W1601,I0021,I0020,W1618,W1619,W1630,W1631,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0142,W0704,W1628,W1629,W1608,I0011,C0303,C1001
```
elbaschid: Are these the default settings for ``disable``? Seems like a whole lot of things are disable here. Can you give a bit more details on the what and why?

python/pylintrc
```diff
@@ -0,0 +1,371 @@
...
+max-returns=6
+
+# Maximum number of branch for function / method body
+max-branches=12
+
+# Maximum number of statements in function / method body
+max-statements=50
+
+# Maximum number of parents for a class (see R0901).
+max-parents=7
```
kyoung: :+1:

elbaschid:
From how I understand it, this is a setup and docs for running PEP8 checks but we don't run them automatically as part of the test script? I think it would be worth adding that into `circle.yml` to ensure it actually gets run by the developers.

python/pylintrc
benlast: It's in the default... I believe it comes from some library package that uses it as a standard name (possibly in `threading`).

python/pylintrc
benlast: Don't think so! I read this as 'don't use these as names for your own methods'

python/pylintrc
benlast: Fair point... do we have a preference? Personally I like docstrings for almost every function unless the behaviour is super obvious, but we could push this threshold up.

python/pylintrc
benlast: Most of them came from the default, and I spot checked them and saw that they were generally spotting Python3 incompatibilities. I added some to remove things that aren't really relevant (e.g., lines that have whitespace at the ends before the \n).

benlast:
@elbaschid I agree that running the checks automatically would be good, but that's a whole other rabbithole to go down, since we'd need agreement on whether tests fail if there are violations (which would mean portal became undeployable), or if there are new violations (how do we track old and find new), and so on.

python/pylintrc
benlast: You're exactly right - I misunderstood the use of that option. Changed to exclude just `input`. I think the others were deprecated because you can achieve the same result with list comprehensions or generators.

python/pylintrc
cbildfell: this is a tough one, since some lines are more complex than others. 3 might be more reasonable?

dbader:
Thanks for putting this together @benlast! This will give us enough of a foundation to try these tools and figure out where we need to do some more tuning. Btw this is an interesting article that talks about applying this in practice: http://www.jeffknupp.com/blog/2013/11/15/supercharge-your-python-developers/

PR #47, opened 2015-02-19
Update readme.md

ellekasai:
The link did not work. So I fixed it.


jeffkamo:
Good catch, thanks @ellekasai!

PR #48, opened 2015-02-27
scss-lint 0.34.0

cole-sanderson:
scss-lint 0.34.0, new and improved scss-lint:

- BangFormat: Checks spacing of ! declarations, like !important and !default

- ColorVariable: Ensures color literals are used only in variable declarations.

- ImportantRule: Reports the use of !important in properties.

- ImportPath: Checks formatting of the basenames of @imported partials

- Indentation: Checks for consistent indentation of nested declarations and rule sets.

- NestingDepth: Checks for rule sets nested deeper than a specified maximum depth.

- PropertyCount: Checks that the number of properties in a rule set is under a defined limit.

- QualifyingElement: Checks for element selectors qualifying id, classe, or attribute selectors.

- SelectorFormat: Checks that selector names use a specified convention

- SingleLinePerProperty: Checks that all properties in a rule set are on their own distinct lines.

- SpaceAfterPropertyName: Report presence of whitespace between property name and colon?

- TrailingZero: Checks for unnecessary trailing zeros in numeric values with decimal points.

- UnnecessaryParentReference: Checks for unnecessary uses of the parent reference (&) in nested selectors.

- VariableForProperty: Reports the use of literals for properties where variables are preferred.

- VendorPrefixes: Checks for vendor prefixes.

- PropertySortOrder: new and improved, now support seperate groups

### DIFF
--- a/css/.scss-lint.yml
+++ b/css/.scss-lint.yml
@@ -6,16 +6,24 @@ exclude: [
 
 # Default application configuration that all configurations inherit from.
 linters:
+    # Checks spacing of ! declarations, like !important and !default
+    BangFormat:
+        enabled: true
+        space_before_bang: true
+        space_after_bang: false
+
     # Prefer border: 0 over border: none?
+    # Configure: `zero` or `none`
     BorderZero:
         enabled: true
-
-    # Should selectors be all lowercase?
-    CapitalizationInSelector:
-        enabled: false
+        convention: zero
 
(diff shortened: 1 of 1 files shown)


yourpalsonja:
+1. Let's merge this–it will stop `should be written in lowercase with hyphens` errors when linting.

jeffkamo:
Hmm, I'm surprised we left this unmerged for as long as we did.

Agreed, +1

PR #51, opened 2015-03-26
Also can use PascalCase for modules

jansepar:
Status: **Opened for visibility**

Reviewers: Any Mobify engineer!

## Changes
- Allow the use of PascalCase for modules

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -255,12 +255,18 @@ var isVisible = true;
 
 ##Use PascalCase only for constructors or modules
 
+
+### Constructors
+
 ````javascript
 // bad
 var Router = new Router();
 ````
 
-### Constructors
+````javascript
+// bad
+var router = new Router();
+````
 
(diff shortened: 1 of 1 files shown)


tedtate:
:+1:

ry5n:
Is there a precedent for PascalCase modules?

tedtate:
@ry5n precedent in our code? I think for the most part this is the standard that we use, we are just codifying it officially here.

ry5n:
I mean precedent in the JS community. I’m not the best person to critique the decision, but if this is a Mobify-specific idiom I’m not sure it’s a great idea: it could confuse people used to the “normal” meaning of PascalCase.

tedtate:
I've seen it written both ways. In many cases modules are really more like namespaces where I think the leading upper-case is a nice.

ry5n:
Thanks, good to know.

PR #54, opened 2015-04-21
[WORK IN PROGRESS] Python style guide

dbader:
@elbaschid and I want to put together a style guide for Python, modelled after the style guide for JavaScript. 

This is a work in progress. Feel free to jump in/add comments! :)

### DIFF
--- a/python/styleguide.md
+++ b/python/styleguide.md
@@ -1,3 +1,108 @@
 # Python Code Style Guide
 
-Hello, World!
+
+Documentation
+-------------
+
+* We use doc strings for package, module, class and function documentation.
+    * A doc string is a string that is the first statement in a package, module, class or function.
+    * These strings can be extracted automatically through the __doc__ member of the object. 
+
+* We always use the three double-quote `"""` format for doc strings.
+    * This is defined in [PEP 257](https://www.python.org/dev/peps/pep-0257/)
+
+* A doc string should be organized as
+    * a summary line (one physical line),
+    * a blank line,
+    * the rest of the doc string.
+
+### On Block comments
+
+* Comment on complicated parts of the code
+    * If you're going to have to explain it at the next code review, you should comment it now.
+
+* Don't describe the code.
+
+
+### Example
+
+* Function or method comments: 
+```python
+def fetch_bigtable_rows(big_table, keys, other_silly_variable=None):
+    """
+    Fetches rows from a Bigtable.
+
+    Retrieves rows pertaining to the given keys from the Table instance
+    represented by big_table.  Silly things may happen if
+    other_silly_variable is not None.
+
+    Args:
+        big_table: An open Bigtable Table instance.
+        keys: A sequence of strings representing the key of each table row
+            to fetch.
+        other_silly_variable: Another optional variable, that has a much
+            longer name than the other args, and which does nothing.
+
+    Returns:
+        A dict mapping keys to the corresponding table row data
+        fetched. Each row is represented as a tuple of strings. For
+        example:
+
+        {'Serak': ('Rigel VII', 'Preparer'),
+         'Zim': ('Irk', 'Invader'),
+         'Lrrr': ('Omicron Persei 8', 'Emperor')}
+
+        If a key from the keys argument is missing from the dictionary,
+        then that row was not found in the table.
+
+    Raises:
+        IOError: An error occurred accessing the bigtable.Table object.
+    """
+    pass
+```
+
+* Class comments:
+```python
+class SampleClass(object):
+    """Summary of class here.
+
+    Longer class information....
+    Longer class information....
+
+    Attributes:
+        likes_spam: A boolean indicating if we like SPAM or not.
+        eggs: An integer count of the eggs we have laid.
+    """
+
+    def __init__(self, likes_spam=False):
+        """Inits SampleClass with blah."""
+        self.likes_spam = likes_spam
+        self.eggs = 0
+
+    def public_method(self):
+        """Performs operation blah."""
+```
+
+* Block or inline comments:
+
+```python
+# We use a weighted dictionary search to find out where i is in
+# the array.  We extrapolate position based on the largest num
+# in the array and the array size and then do binary search to
+# get the exact number.
+
+if i & (i-1) == 0:        # true iff i is a power of 2
+```
+
+
+### Reason(s)
+
+* Documentation style is formalized in [PEP-257](https://www.python.org/dev/peps/pep-0257/) and a community standard.
+* Tools like [Sphinx](http://sphinx-doc.org/index.html) can extract *docstrings* and auto-generate documentation.
+* The Google style formatting is very readable in the source code compared to other styles.
+
+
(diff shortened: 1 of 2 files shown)


python/styleguide.md
```diff
@@ -0,0 +1,3 @@
+# Python Code Style Guide
+
+Hello, World!
```
jansepar: I disagree. Hello World should not be the only thing I'm allowed to put into my code.

:)

elbaschid:
Here's a list of existing styleguides that are worth looking at as a reference:

* ipython: http://nbviewer.ipython.org/github/addfor/tutorials/blob/master/python-ipython/py04v03_python_style_guide.ipynb
* Google: https://google-styleguide.googlecode.com/svn/trunk/pyguide.html
* http://docs.python-guide.org/en/latest/writing/style/
* Mozilla: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style
* Idiomatic Python: http://python.net/~goodger/projects/pycon/2007/idiomatic/handout.html

elbaschid:
I also just found this interesting list of "anti-patterns" collected for Python with the rationale behind why they are bad or sub-optimal:

http://docs.quantifiedcode.com/python-anti-patterns/index.html

python/styleguide.md
```diff
@@ -0,0 +1,108 @@
+# Python Code Style Guide
+
+
+Documentation
+-------------
+
+* We use doc strings for package, module, class and function documentation.
+    * A doc string is a string that is the first statement in a package, module, class or function.
+    * These strings can be extracted automatically through the __doc__ member of the object.
```
dbader: Markdown screws up the formatting here because it parses `__` as *bold-this-word*,  `__doc__` should work.

elbaschid: damn you markdown :)

python/styleguide.md
```diff
@@ -0,0 +1,108 @@
...
+    * A doc string is a string that is the first statement in a package, module, class or function.
+    * These strings can be extracted automatically through the __doc__ member of the object. 
+
+* We always use the three double-quote `"""` format for doc strings.
+    * This is defined in [PEP 257](https://www.python.org/dev/peps/pep-0257/)
+
+* A doc string should be organized as
+    * a summary line (one physical line),
+    * a blank line,
+    * the rest of the doc string.
```
dbader: Idea: We should give some examples of what *good* docstrings look like. I personally like what PEP 257 says about that:

> * The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

> * The one-line docstring should NOT be a "signature" reiterating the function/method parameters (which can be obtained by introspection). [...]

python/styleguide.md
```diff
@@ -0,0 +1,108 @@
...
+    Raises:
+        IOError: An error occurred accessing the bigtable.Table object.
+    """
+    pass
+```
+
+* Class comments:
+```python
+class SampleClass(object):
+    """Summary of class here.
```
dbader: Historically most of our Python code-bases used a docstring format like this (even for short comments):

```
"""    <--- triple quotes on their own line
The comment.
       <--- blank line
"""
```

It might make sense to continue that practice because consistency :rainbow:. At some point this was also written down in Confluence but I can't find the page at the moment. Then again it might make sense to just adopt something like PEP 257 so we don't have to come up with our own formatting rules :)

python/styleguide.md
```diff
@@ -0,0 +1,108 @@
...
+
+* Block or inline comments:
+
+```python
+# We use a weighted dictionary search to find out where i is in
+# the array.  We extrapolate position based on the largest num
+# in the array and the array size and then do binary search to
+# get the exact number.
+
+if i & (i-1) == 0:        # true iff i is a power of 2
```
dbader: Our current pylint settings will ask us to convert `(i-1)` to `(i - 1)` (add spaces)

python/styleguide.md
```diff
@@ -0,0 +1,108 @@
+# Python Code Style Guide
+
+
+Documentation
+-------------
```
dbader: Meta-style comment about the Markdown style we're using here in this Python style guide: Should we standardize on using `##` for `h2` headings etc.? I personally always get confused what heading level underlining with `----` or `====` produces :-)

dbader:
Thanks for kicking things off with the section on docs! Great job @elbaschid!

python/styleguide.md
```diff
@@ -0,0 +1,108 @@
...
+    * a summary line (one physical line),
+    * a blank line,
+    * the rest of the doc string.
+
+### On Block comments
+
+* Comment on complicated parts of the code
+    * If you're going to have to explain it at the next code review, you should comment it now.
+
+* Don't describe the code.
```
dbader: ... Don't describe the code and avoid superfluous comments. Generally it's best to convey that information *in the actual code*, for example by using appropriate naming:

Harmful:

```python
d = 0  # Elapsed time in days
```

Better:

```python
elapsed_time_in_days = 0
# or
days_elapsed = 0
```

dbader:
Summarizing the discussion @elbaschid and I had today:

* We need to approach this style guide in a slightly different way, otherwise we'll run the danger of writing a summary of Clean Code and other books.

* The approach we decided to explore is to split things up into  a "Python Style Guide" and (maybe later) a more general "Developer's Guide to Writing Code at Mobify".

* Things that should go into the **Python Style Guide**:
    * A small set of specific rules around how Python code should be formatted and what the community considers idiomatic.
    * These would be "strong guidelines" or "rules" that we expect all (future) Python code at Mobify to follow.
    * (It would be great to find a way to automate most of the checking)
    * Example: Google's code style guide

* Things that go into the "Developer's Guide":
    * Best practices that the dev community agrees upon
    * An aggregation of reference material, like books and articles.
    * This is a more "inspirational" document that applies to all projects and languages.
    * It would also serve as a basis for discussion in code reviews

This will allow us to finish a first draft of the style guide with much reduced scope.

benlast:
This is a great effort, and I genuinely applaud the effort going into it, but I worry that it's way too easy to turn a "style guide" into a long exhaustive (and exhausting) list of every bit of "best practice" that can be found, which then leads to lawyer-esque reviews, where anything that can be considered as a violation of any of the many rules is called out in comments. Is that really how we want reviews to work?

For example: the section of 'naming' in the current commits - it's well-intentioned, but do we really need to spell out all the different aspects of how names should be chosen? Are all these rules needed? Is a reviewer supposed to check for compliance?

"A small set of specific rules around how Python code should be formatted"... what's wrong with PEP8? Are we **really** going to spend time looking at formatting?

"what the community considers idiomatic", and "best practices that the dev community agrees upon" - what if the 'community' doesn't agree? Do we decide by majority vote?

I think there should be a clear distinction between:
* rules that are mandatory (which should be automatically enforceable by tools, not by humans)
* general guidelines on how we write code - for enlightenment, discussion, guidance of new developers, etc.

PR #57, opened 2015-06-04
updating self documenting selectors guide based on UI dev meeting dis…

yourpalsonja:
Status: **Ready for Review**
Owner: @yourpalsonja
Reviewers: @jeffkamo @ry5n @avelinet @nastiatikk @cole-sanderson @mlworks

## Changes
- Added some examples to the *Component modifiers that affect subcomponents* documentation
- Added suggestions for commenting that could be used with different methods

## Todos:
- [x] Developer +1@ry5n
- [x] Another Developer +1 @avelinet

### Notes:
These changes are based on our discussion on [May, 27 2015](https://docs.google.com/document/d/1eZCqylKXcVxPA7iWxXRHfLboFHyQJmx80u7CS-dRUc4/edit#heading=h.m7s0tjuitd2x)

### DIFF
--- a/css/class-naming-conventions/readme.md
+++ b/css/class-naming-conventions/readme.md
@@ -112,7 +112,7 @@ These are used to modify components or subcomponents. They are always chained to
 
 ### Component modifiers that affect subcomponents
 
-Sometimes a component modifier will affect its sub-components. There are a few methods you can use to accomplish this. Try to pick one and stick to it throughout the project, adding comments as needed.
+Sometimes a component modifier will affect its sub-components. There are several methods you can use to accomplish this. As much as possible, stick to one method in your project.
 
 ```html
 <div class="c-blog-post c--featured">
@@ -124,10 +124,10 @@ Sometimes a component modifier will affect its sub-components. There are a few m
 ```
 
 
-#### 1. Nested inside the component
+#### 1. Styles grouped with modifier
 Nest the `.c-component__sub-component` elements inside the `.c-component` SCSS.
 
-This method works well when a modifier requires simple changes to a small number of sub-components.
+This method allows you to quickly update or edit the styles for all elements affected by a modifier.
 
 ```scss
 .c-blog-post {
@@ -141,11 +141,8 @@ This method works well when a modifier requires simple changes to a small number
 }
 ```
 
+*or*
 
-#### 2. Root level as a selector chain
-Use a selector chain with `.c-component.c--modifier`, and nest the `.c-component__sub-component` elements within it.
-
-This method works especially well when a component has multiple modifiers or state classes that can be coupled with it.
 
 ```scss
 .c-blog-post.c--featured {
@@ -157,23 +154,22 @@ This method works especially well when a component has multiple modifiers or sta
 }
 ```
 
-When using the above methods in larger files, add a comment to the `.c-component__sub-component` like so:
+In larger files, adding a comment in the `.c-component__sub-component` notes can be helpful:
 ```scss
 // Blog Post Title
 // ---
 //
-// Modifier styles can be found in:
-// .c-blog-post.c--featured
+// Modified by .c-blog-post.c--featured
 
 .c-blog-post__title {
     ...
 }
 ```
 
-#### 3. Nested inside the sub-component
-Nest the modifier code inside the subcomponent using `.c-component.c--modifier &`.
+#### 2. Styles grouped with sub-component
+Nest the modifier code inside the sub-component using `.c-component.c--modifier &`.
 
-This method works well when a root-level modifier affects multiple sub-components, or when you are working with complicated/intricate sub-components.
+This method makes it easier to visualize the differences between a sub-component and its modified states.
 
 ```scss
 .c-blog-post__title {
@@ -185,7 +181,7 @@ This method works well when a root-level modifier affects multiple sub-component
 }
 ```
 
-When using the above method in larger files, add a comment to the `.c-component` like so:
+In larger files, adding a comment in the `.c--modifier` notes can be helpful:
 ```scss
 // Blog Post
 // ===
@@ -195,8 +191,7 @@ When using the above method in larger files, add a comment to the `.c-component`
     // Featured Post
     // ---
     //
-    // Also modifies:
-    // .c-blog-post__title
+    // Also modifies .c-blog-post__title
 
     .c--featured {
         ...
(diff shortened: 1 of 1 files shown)


css/class-naming-conventions/readme.md
```diff
@@ -123,21 +123,80 @@ Sometimes you'll write a modifier for a component and you want that modifier to
...
+    ...
+
+    .c-blog-post__title {
+        ...
+    }
+}
+```
+
+When using the above methods in larger files, add a comment to the `c-component__sub-component` like so:
+```
```
jeffkamo: Might be worth including SCSS syntax highlighting here.

css/class-naming-conventions/readme.md
```diff
@@ -123,21 +123,80 @@ Sometimes you'll write a modifier for a component and you want that modifier to
...
     .c-blog-post.c--featured & {
-        [STYLES]
+        ...
     }
 }
 ```
 
-This might look a little weird at the outset but it's the best way to ensure that all of a components styles stay in the same place. It also ensures that no modifier styles are accidentally inherited where they shouldn't be.
+When using the above method in larger files, add a comment to the `c-component` like so:
+```
```
jeffkamo: Might be worth including SCSS syntax highlighting here too.

css/class-naming-conventions/readme.md
```diff
@@ -123,21 +123,80 @@ Sometimes you'll write a modifier for a component and you want that modifier to
...
+
+        .c-blog-post__title {
+            ...
+        }
     }
 }
+```  
+
+
+#### Below the component as a selector chain
```
jeffkamo: Would it be more clear to say "Root Level" instead of "Below the component"?

css/class-naming-conventions/readme.md
```diff
@@ -123,21 +123,80 @@ Sometimes you'll write a modifier for a component and you want that modifier to
 </div>
 ```
 
+
+#### Nested inside the component
```
avelinet: Would it be clearer if the headings for the different methods were numbered 1, 2, 3?

avelinet: Also since we are mentioning there are a few different methods, would it be helpful to include a little blurb on which types of situation one method might be preferable over another? I know we did talk about that a bit in our meeting.

jeffkamo: Probably a good idea, yeah

yourpalsonja:
Thanks all! Gonna update this afternoon. :sun:

css/class-naming-conventions/readme.md
yourpalsonja: Great idea. :dancer:

yourpalsonja:
Do you think we could add some best practices around where modifiers should be placed?

Example from the [meeting](https://docs.google.com/document/d/1eZCqylKXcVxPA7iWxXRHfLboFHyQJmx80u7CS-dRUc4/edit#bookmark=id.xp0r8pt9j7t) (slightly paraphrased for clarity):

> If entire component is changing state, have the modifier live on the root because with the way partials work in dust, it’s easier to have modifier on the root than on sub-components.

I don't know if this should be added here, or elsewhere in the document. Thoughts?

ry5n:
@yourpalsonja Hmmm… not sure if that should be a rule that we document. I’m not against documenting rules of thumb, but we already have a lot of guidelines for CSS and this might not be ready to include yet.

yourpalsonja:
@ry5n Cool. I don't think it should live in the CSS portion of the guide, really–maybe HTML? 

It was something that came up in the meeting that I don't remember reading about in our documentation, but is a solid rule to follow. Perhaps a topic for a future meeting. :)

avelinet:
:+1:

css/class-naming-conventions/readme.md
```diff
@@ -123,21 +123,86 @@ Sometimes you'll write a modifier for a component and you want that modifier to
...
+}
+```
+
+When using the above methods in larger files, add a comment to the `.c-component__sub-component` like so:
+```scss
+// Blog Post Title
+// ---
+//
+// Modifier styles can be found in:
+// .c-blog-post.c--featured
```
ry5n: For small and medium-sized files (ideally most of them) I doubt the cross-reference is needed, so as long as this is a recommendation rather than a requirement, I’d cool with it.

yourpalsonja: @ry5n good point, I'll make it more obvious that this is a suggestion, not a rule.

ry5n:
I’m not sure I understand the three distinctions being made here. I thought that we had made a distinction only between two ways of writing out modifiers (see below). Am I missing something?

Grouping by modifier: all subcomponents that change with the modified component
```scss
.c-component.c--modifier {
    .c-component__subcomponent {…}
}

.c-component.c--other {
    .c-component__subcomponent {…}
}
```

Grouping by sub-component: all styles (including modifications) are kept together
```scss
.c-component__subcomponent {
    .c-component.c--modifier & {
        …
    }

    .c-component.c--other & {
        …
    }
}
```

yourpalsonja:
@ry5n I'm breaking down the first method you mentioned in your example.

We talked about having the "grouping by modifier" method that you mentioned, and there are two ways that can be achieved: breaking out the component and modifier into a chain at the root level(like you demonstrated above, and is No. 2 in the examples), or keeping the modifiers nested inside the component declaration block (No. 1 in the examples).

```scss
.c-component {
    ...
    
   &.c--modifier {
       ...
       .c-component__subcomponent { ... }
    }
}
```
The above method works well with a component that has maybe 1 modifier, and the modifier affects 1 or 2 sub-components, each requiring a few short lines of code.

Hope that clarifies my thought process a bit.

Any other thoughts on whether we should nix method No. 1 altogether, and use No. 2 as a best practice when grouping by modifier?

ry5n:
Oh! I see. Sorry, you’re right. I’m not sure we need to differentiate no.1 and no.2 though, and I don’t think it’s a big deal which way they are written out. I think it’s probably easier to understand if we focus on the difference between grouping by modifier or by subcomponent, since I think that is the key issue here. I’d also suggest sticking with one example for a combined 1 & 2, whichever you think is clearer.

yourpalsonja:
@ry5n I combined both 1 & 2 as per your suggestion–it makes it a lot clearer. I left both code examples in, though.

The extra information might be helpful in this case, based on my thoughts about my initial ramp up (so long ago now :hourglass:  :smile_cat:).

ry5n:
@yourpalsonja Excellent! Yeah, this is much clearer. Keeping both examples is also a good call. +1.

yourpalsonja:
:tada:

ry5n:
🍰

PR #59, opened 2015-06-11
Updated scss-lint to 0.39.0

cole-sanderson:
Updated scss-lint to 0.39.0

Status: Ready for review

Reviewers: @jeffkamo @avelinet @mlworks @yourpalsonja @ry5n

## Changes
- BemDepth, checks for BEM selectors with more elements than a specified maximum number.
- ExtendDirective, checks that `@extend` is never used.
- NameFormat, removed 'BEM' as configure.
- PropertyUnits, check for allowed units.
- TrailingWhitespace, checks for trailing whitespace on a line.
- PropertySortOrder, added min_properties.

### DIFF
--- a/css/.scss-lint.yml
+++ b/css/.scss-lint.yml
@@ -14,7 +14,7 @@ linters:
 
     # Checks for BEM selectors with more elements than a specified maximum number.
     BemDepth:
-        enabled: false
+        enabled: true
         max_elements: 1
 
     # Prefer border: 0 over border: none?
@@ -290,7 +290,7 @@ linters:
     PropertySortOrder:
         enabled: true
         ignore_unspecified: false
-        min_properties: 1
+        min_properties: 2
         order:
             - content
             - quotes
--- a/css/.scss-lint.yml
+++ b/css/.scss-lint.yml
@@ -12,6 +12,11 @@ linters:
         space_before_bang: true
         space_after_bang: false
 
+    # Checks for BEM selectors with more elements than a specified maximum number.
+    BemDepth:
+        enabled: false
+        max_elements: 1
+
     # Prefer border: 0 over border: none?
     # Configure: `zero` or `none`
     BorderZero:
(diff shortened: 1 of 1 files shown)


css/.scss-lint.yml
```diff
@@ -12,6 +12,11 @@ linters:
         space_before_bang: true
         space_after_bang: false
 
+    # Checks for BEM selectors with more elements than a specified maximum number.
+    BemDepth:
+        enabled: false
```
ry5n: Can we turn this on? Would be a helpful check for folks just just starting with BEM.

css/.scss-lint.yml
```diff
@@ -265,6 +290,7 @@ linters:
     PropertySortOrder:
         enabled: true
         ignore_unspecified: false
+        min_properties: 1
```
ry5n: The default for `min_properties` is 2, which kinda makes sense since sorting isn’t necessary for one-rule declarations. Maybe this should stay at `2`?

ry5n:
Thanks for taking the time to update! I rather like that we are setting all the things here, even if we keep the defaults: it’s explicit. Thumbsup.

For other reviewers, these are the settings I had to look up:

- min_properties: Number of rules in a ruleset required for sorting to be enabled for that ruleset
- BemDepth.max_elements: Number of sub-components allowed in BEM classnames. We don’t allow more than one, e.g. this is bad: `component_sub_subsub {…}`

yourpalsonja:
@cole-sanderson This is great! So excited that variable names aren't going to give me linting errors anymore. :tada:

css/.scss-lint.yml
cole-sanderson: Yeah good point, I ll change to 2.

css/.scss-lint.yml
cole-sanderson: Sure, I ll turn that on. So if we write `.block__element__subelement` and we will get error.

cole-sanderson: Or should we allow max 2 elements or keep to 1 element?

css/.scss-lint.yml
cole-sanderson: Nevermind, I just read your message about the BemDepth.

ry5n:
+1 from me!

jeffkamo:
Looks good to me too! +1

PR #64, opened 2015-08-31
Adding a few new rules based on our book club meeting!

jansepar:


donnielrt:
:+1: to both changes, surprised we didn't already have these!

javascript/README.md
```diff
@@ -88,6 +88,21 @@ var length = items.length;
 var name = 'foo';
 ````
 
+##Use semi-colons
+
+There are many more ways that things can break *without* semi-colons then *with* semi-colons.
```
jeremywiebe: s/then/than/ (grammar police, sorry... :) )

kschmidtdev: :+1:

PR #67, opened 2015-09-30
Remove rule preventing extra parens

MikeKlemarewski:
Up for debate!

### DIFF
--- a/javascript/.eslintrc
+++ b/javascript/.eslintrc
@@ -93,9 +93,6 @@
         // Enforce single quotes for string literals
         "quotes": [2, "single"],
 
-        // Disallow unnecessary parentheses
-        "no-extra-parens": 2,
-
         // Disallow unnecessary semicolons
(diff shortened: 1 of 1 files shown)


ry5n:
+1

marlowpayne:
Rule doc [here](http://eslint.org/docs/rules/no-extra-parens)

Maybe we should set the rule to `functions`, so function expressions will still be checked for extra parens?
e.g. `((function foo() {}))();`

haroldtreen:
Ya, I think the functions case is still good. The times this rule has tripped me has only been in expressions.

MikeKlemarewski:
@marlowpayne I like that idea. How do I set it to still check functions?

marlowpayne:
@MikeKlemarewski `"no-extra-parens": [2, "functions"]`

marlowpayne:
:+1:

PR #69, opened 2015-10-01
Remove parens rule for functions

MikeKlemarewski:
Remove parens rule completely as it had some quirky behaviour.


marlowpayne:
:+1:

PR #70, opened 2015-10-05
Notes from Effective JavaScript Chapter 5

jeremywiebe:
Adds one new rule based on book club discussion.
  
Reviewers: @jansepar @spiside @noahadams @fractaltheory @haroldtreen 
~~JIRA: N/A~~
~~Linked PRs: N/A~~
  
## Changes
- Add rule "Prefer `[].forEach.call(some_array_like_object);` over `Array.prototype.forEach.call(some_array_like_object);`."
  
## How to test-drive this PR
- Read the docs
  
## Research resources
- http://jsperf.com/foreach-vs-array-prototype-foreach

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -382,3 +382,22 @@ if ("0" == false) {console.log('foo')} // this will console log! bad!
 // good
 if ("0" === false) {console.log('foo')}
 ```
+
+##Calling array methods on array-like objects
+
+Use functions on Array.prototype on array-like objects. (eg. 'arguments' is not an array, but 
+is array-like).
+
+###Prefer [].forEach(obj) over Array.prototype.forEach(obj)
+
+Both `[].forEach` and `Array.prototype.forEach` resolve to the same function. Performance
+between the two is almost identical. Prefer the short, more succinct version.
+
+```javascript
+// Good
+[].slice.call(arguments); 
+
+// Bad!
+Array.prototype.slice.call(arguments);
+```
+
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -390,7 +390,9 @@ is array-like).
 
 ###Prefer using methods from `[]` over `Array.prototype`
 
-[Performance between the two styles is almost identical](http://jsperf.com/foreach-vs-array-prototype-foreach). Prefer the short, more succinct version.
+Prefer the short, more succinct version.
+
+[Performance between the two styles is almost identical](http://jsperf.com/foreach-vs-array-prototype-foreach). 
 
 ```javascript
 // Good
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -388,10 +388,9 @@ if ("0" === false) {console.log('foo')}
 Use functions on Array.prototype on array-like objects. (eg. 'arguments' is not an array, but 
 is array-like).
 
-###Prefer [].forEach(obj) over Array.prototype.forEach(obj)
+###Prefer using methods from `[]` over `Array.prototype`
 
-Both `[].forEach` and `Array.prototype.forEach` resolve to the same function. Performance
-between the two is almost identical. Prefer the short, more succinct version.
+Performance between the two styles is almost identical. Prefer the short, more succinct version.
 
 ```javascript
 // Good
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -385,7 +385,7 @@ if ("0" === false) {console.log('foo')}
 
 ##Calling array methods on array-like objects
 
-Use functions on Array.prototype on array-like objects. (eg. 'arguments' is not an array, but 
+Use functions on Array.prototype for array-like objects. (eg. 'arguments' is not an array, but 
 is array-like).
 
 ###Prefer using methods from `[]` over `Array.prototype`
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -390,7 +390,7 @@ is array-like).
 
 ###Prefer using methods from `[]` over `Array.prototype`
 
-Performance between the two styles is almost identical. Prefer the short, more succinct version.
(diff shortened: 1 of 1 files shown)


javascript/README.md
```diff
@@ -382,3 +382,21 @@ if ("0" == false) {console.log('foo')} // this will console log! bad!
...
 ```
+
+##Calling array methods on array-like objects
+
+Use functions on Array.prototype for array-like objects. (eg. 'arguments' is not an array, but 
+is array-like).
+
+###Prefer using methods from `[]` over `Array.prototype`
+
+Performance between the two styles is almost identical. Prefer the short, more succinct version.
```
jansepar: Could we link to the js perf thing we looked at during our meeting?

javascript/README.md
```diff
@@ -382,3 +382,21 @@ if ("0" == false) {console.log('foo')} // this will console log! bad!
...
+
+```javascript
+// Good
+[].slice.call(arguments);
```
jvaill: Hm. We might want to change this to something like `[].slice.call(this, arguments...);`. The first argument to call is the value of this to bind the function to, not the arguments that are actually passed to `slice`.

jvaill:
Do we always prefer `[].forEach.call(this, args)` versus `[].forEach(args)`? Do we always want to be explicit about the value of this being bound?

MikeKlemarewski:
I'm in favour of readability and quite often lean towards longer bits of code if they are more clear.
I think `[].forEach` should be just as readable as `Array.prototype.forEach` for almost every engineer at Mobify. If you write JavaScript, you should know about the prototype chain and know that these statements are equivalent.

javascript/README.md
jeremywiebe: It's linked in the PR, but I'll pull it down into the README

jvaill:
Actually, I'm going to answer my first question:

The docs in this PR make it seem like this is the style we favour:
```javascript
var a = ['h', 'e', 'y'];
[].forEach.call(a, function(ele) { console.log(ele); });
```

Whereas it sounds like the style we're proposing here is:
```javascript
var a = ['h', 'e', 'y'];
a.forEach(function(ele) { console.log(ele); });
```

Is that correct?

javascript/README.md
haroldtreen: Without `this` being passed, forEach won't know which object it's iterating over.

jeremywiebe:
@jvaill no, the intent was to favour `[].forEach.call(arguments, function(arg) { console.log(arg); }` over `Array.prototype.forEach.call(arguments, function(arg) { console.log(arg); }`

MikeKlemarewski:
This is when calling Array methods on Array-like objects. When calling methods on actual arrays you'd still use `myArray.forEach()`

jansepar:
I think this was to specifically address that if you want to use a method on a prototype (so, say you want to call an array method on an array-like object), then grab it directly by doing `[].prototype` rather than doing `Array.prototype`.

If we are actually accessing methods on an instance, the preference is still to access directly on that instance. Sounds like we need to make this more clear though if this was something @jvaill was confused about.

jvaill:
@jeremywiebe Ahhhh. I thought both of the headings you added were at the same level, I missed that `Calling array methods on array-like objects` is a sub-heading by looking at the diff.

Y'all can just ignore me. This makes sense now!

jvaill:
@jansepar sounds good, this makes perfect sense now.

I think it's probably clear enough, I just missed that the second heading was a sub-heading of the first in the diff (I thought they were two independent points). It should be clearer once this is merged and rendered as markdown. :) Thanks for the clarification.

javascript/README.md
jeremywiebe: Just tried it, and that doesn't seem to work.  I believe the correct syntax is to pass `arguments` as the first call, as it's "array-like".
```
> var x = function() {
... var y = [].slice.call(arguments);
... console.log("Result: " + y);
... }
undefined
> x(29392,2,2,2)
Result: 29392,2,2,2
undefined
> var x = function() {
... var y = [].slice.call(null, arguments);
... console.log("Result: " + y);
... }
undefined
> x(2,3,4)
TypeError: Array.prototype.slice called on null or undefined
    at x (repl:2:18)
    at repl:1:1
    at REPLServer.defaultEval (repl.js:164:27)
    at bound (domain.js:250:14)
    at REPLServer.runBound [as eval] (domain.js:263:12)
    at REPLServer.<anonymous> (repl.js:393:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
    at REPLServer.Interface._line (readline.js:549:8)
```

spiside:
Sigh 😔 I admit defeat. I'll 👍 the `[].forEach.call(...)` over the `Array.prototype`

javascript/README.md
```diff
@@ -382,3 +382,23 @@ if ("0" == false) {console.log('foo')} // this will console log! bad!
...
+
+Prefer the short, more succinct version.
+
+[Performance between the two styles is almost identical](http://jsperf.com/foreach-vs-array-prototype-foreach). 
+
+```javascript
+// Good
+[].forEach.call(arguments, function(arg) { console.log(arg); });
+
+// Bad!
```
spiside: I don't wanna to say this is Bad! though. Maybe say something like `Longer`

jansepar: @spiside it's just "Bad" because if you are doing it that way, you are not following the style guide :P. Also we use the terminology "Good" and "Bad" for all of our examples for each point in this document.

jeremywiebe: @spiside I tried to soften the blow by saying "Prefer" in the title. :-)

PR #71, opened 2015-10-08
Update Swift code style guide to include some Swift 2.0 stuff

jvoll:
Guard is awesome, let's use it when possible!

@kschmidtdev, @jeremywiebe, and I agreed to leave error handling as is, at least for now, since we think it works better for our highly async code base. Feedback welcome!

### DIFF
--- a/swift/README.md
+++ b/swift/README.md
@@ -177,7 +177,6 @@ guard let x = x else {
 }
 ```
 
-
 Objective-C interop
 ===================
 
--- a/swift/README.md
+++ b/swift/README.md
@@ -37,7 +37,7 @@ class Component {
 How do I handle errors in Swift?
 ================================
 
-As of Swift 2.0 exceptions have been introduced into the language and many of the built-in APIs make use of them. We have decided to stick with our existing pattern of returning errors.
+As of Swift 2.0 exceptions have been introduced into the language and the APIs that formerly used `someMethod(..., error: *NSError)` make use of them. We have decided to stick with our existing pattern for returning errors.
 
 You have a couple options:
 
--- a/swift/README.md
+++ b/swift/README.md
@@ -154,6 +154,30 @@ Place the closing `}` on the same line if the body of the block if empty:
 let callback: RpcMethodCallback = { result in }  // Do nothing
 ```
 
+Place bodies of a block on a new line
+=====================================
+
+No single line `if...else` statements
+
+```swift
+// bad
+if condition { return }
+
+// good
+if condition {
+    return
+}
+
+// bad
+guard let x = x else { return }
+
+// good
+guard let x = x else {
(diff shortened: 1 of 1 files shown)


swift/README.md
```diff
@@ -37,7 +37,7 @@ class Component {
 How do I handle errors in Swift?
 ================================
 
-Swift doesn't have exceptions, so errors must somehow be returned to your caller.
+As of Swift 2.0 exceptions have been introduced into the language and many of the built-in APIs make use of them. We have decided to stick with our existing pattern of returning errors.
```
jeremywiebe: Can we clarify that the APIs that use this are all the ones that used to use `someMethod(..., error: *NSError)` ?  People coming from Cocoa will be able to identify with that easily, I think.

swift/README.md
```diff
@@ -249,6 +249,44 @@ if let message = message {
...
+```
+
+```swift
+// Good (Swift 2.0 way using guards)
+func myFunc(optionalKey: String?) {
+    guard let key = key else {
+        // log an error
+        return
+    }
+    guard let queue = dict[key] else { return }
```
jeremywiebe: I know we're trying to be succinct here, but I think our code style also doesn't allow for these single-line blocks. :-)

swift/README.md
```diff
@@ -249,6 +249,44 @@ if let message = message {
...
+    for queueMessage in queue {
+        guard let address = queueMessage["address"] as? MessageAddress else { continue }
```
jeremywiebe: Oh, cool.  Didn't know you could use guard to short-circuit a loop!

swift/README.md
jvoll: Looking through it now I don't actually see us explicitly saying "no" to this. I thought you were right...but I don't see it. For empty closures we allow it to be on the same line but that's the closest thing I can see.

kschmidtdev:
:+1:

swift/README.md
```diff
@@ -249,6 +273,50 @@ if let message = message {
...
+```
+
+```swift
+// Good (Swift 2.0 way using guards)
+func myFunc(optionalKey: String?) {
+    guard let key = key else {
+        // log an error
+        return
+    }
+    guard let queue = dict[key] else {
```
jansepar: Guard, so hot right now. Guard.

swift/README.md
```diff
@@ -154,6 +154,30 @@ Place the closing `}` on the same line if the body of the block if empty:
...
+// bad
+guard let x = x else { return }
+
+// good
+guard let x = x else {
+    return
+}
+```
+
+
```
jeremywiebe: pew pew

jeremywiebe:
One extra line comment but :+1:

PR #79, opened 2015-11-17
Fix lint/comb rule to support both single and double quotes

jeffkamo:
@marlowpayne

At Mobify, we most commonly use both double quotes and single quotes in our CSS. We want to support both!

## Todo

- [x] Figure out how to disable quotes


jeffkamo:
There's a problem: CSScomb doesn't have a way to disable single quotes...

We actually want to support both single and double quotes, but CSScomb won't let us :(

marlowpayne:
Booo! :ghost:

jeffkamo:
Nevermind! We can set the rule to `false` and that works.

marlowpayne:
Looking good! :+1:

PR #81, opened 2015-12-09
Change CSSComb rule to disable leading-zero due to bug

cole-sanderson:
# Change CSSComb rule to disable leading-zero due to bug

Status: **Ready for Review**
Reviewers: @marlowpayne @jeffkamo 

## Changes:
- Change CSScomb rule: disable leading-zero rule due to CSScomb bug

## Bug:
For example, I want to write `margin: $unit*1.5` and with CSSComb, it will rewrite to `margin: $unit*10.5`. It should not add 0 in this case, changes visual big time. 

## How to test:
- cd to mobify-code-style
- `npm link` to update CSSComb rules.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,5 @@
 2.3.6
     - Change CSScomb rule: disable leading-zero rule due to CSScomb bug
-    - Change scss-lint rule: disable LeadingZero rule to work with CSSComb leading-zero rule
 2.3.5
     - Add Sass documentation about placeholders and `@extends`
     - Add CSS documentation on l10n and i18n best practices
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 2.3.6
     - Change CSScomb rule: disable leading-zero rule due to CSScomb bug
+    - Change scss-lint rule: disable LeadingZero rule to work with CSSComb leading-zero rule
 2.3.5
     - Add Sass documentation about placeholders and `@extends`
     - Add CSS documentation on l10n and i18n best practices
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,5 @@
+2.3.6
+    - Change CSScomb rule: disable leading-zero rule due to CSScomb bug
 2.3.5
     - Add Sass documentation about placeholders and `@extends`
     - Add CSS documentation on l10n and i18n best practices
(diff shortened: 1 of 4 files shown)


johnboxall:
By analyzing the blame information on this pull request, we identified @marlowpayne, @jeffkamo and @tedtate to be potential reviewers.

jeffkamo:
@cole-sanderson you also have to update the version in `package.json`

marlowpayne:
:+1: :shipit:

jeffkamo:
We'll want to tweak the linter to disable leading zeros: https://github.com/mobify/mobify-code-style/blob/1a487b4b6dd94a726610e76a5f53e428f4fcf3be/css/.scss-lint.yml#L119

cole-sanderson:
@jeffkamo good catch, we will need to change that to pass scss-lint

marlowpayne:
@jeffkamo @cole-sanderson I don't think we'll need the Sass linter updated, we just don't want CSSComb introducing this bug.

jeffkamo:
Good point, I guess it can work either way.

cole-sanderson:
@marlowpayne we will need to update scss-lint rule or otherwise we will get scss-lint error.

cole-sanderson:
for example in CSSComb it will change from `margin: 0.5em`to `margin: .5em` and if we don't change scss-lint rule and we will get LeadingZero error to change `.5` to `0.5`.

jeffkamo:
Marlow makes a good point though. We still want the linter to catch that formatting error. We're only disabling it in CSSComb because it's introducing a bug. At least the linter will ensure the code is being formatted correctly.

marlowpayne:
What about removing the `leading-zero` rule from CSSComb's config entirely? Then it won't fight us by taking away the leading zeros we want, and our linter will also be happy. :)

yourpalsonja:
That sounds like a good idea, @marlowpayne !

jeffkamo:
We can also set it to `null`

cole-sanderson:
Yeah, we will set it as `null` so leading-rule rule in CSSComb will be ignored.

cole-sanderson:
@jeffkamo @marlowpayne okay now it should be good to "Review" :)

jeffkamo:
+1

marlowpayne:
:+1:

cole-sanderson:
Thank you @marlowpayne and @jeffkamo for review!

tedtate:
Didn't even wait for my :+1: , https://raw.githubusercontent.com/tedtate/gifdump/master/gifs/how-rude-stephanie-full-house.gif

marlowpayne:
YouHaveNoPowerHere @tedtate

PR #82, opened 2016-01-05
Javascript self vs bind(this)

lizcross:
# Self vs bind(this) which do we prefer

Status: **Ready for Review**
Reviewers: @jansepar all js devs

## Changes:

- Took the stance that we prefer self over bind(this)
- Discuss!

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -404,7 +404,6 @@ Array.prototype.forEach.call(arguments, function(arg) { console.log(arg); });
 
 ##Prefer `self` to `bind(this)`
 
-
 ```javascript
 // Good
 AiBot.prototype.answer = function(questionPromise) {
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -445,6 +445,7 @@ UIService.prototype.addEventing = function(eventEmitter) {
         });
     }.bind(this));
 };
+```
 
 ##Method and promise chains
 
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -7,9 +7,9 @@ Please comment code extensively. More comments are always better than fewer comm
 // bad!
 // Shows footer if shopping bag is absent
 var showFooter = !$('.shopping-bag').length;
- 
+
 // good
-// PROJ-12: As per client request, we hide the footer if 
+// PROJ-12: As per client request, we hide the footer if
 // the shopping bag is not visible on the page.
 var showFooter = !$('.shopping-bag').length;
 ````
@@ -37,14 +37,14 @@ In general, always favour comment placement that leads to less indenting.
 Don't commit commented out sections of code back into the repository. Just delete the code. That's what git's history is for!
 
 If a piece of code is very temporarily being removed, and will be reinserted shortly, you might decide to do this anyway. Please leave a detailed comment explaining exactly why. Sorta like those post-its we leave on boxes in the fridge with our name and date so that we know when to chuck 'em out!
- 
+
 
 ##Use single quotes
 
 ````javascript
 // good
 $('.footer')
- 
+
 // bad
 $(".footer")
 ````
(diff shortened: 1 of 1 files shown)


johnboxall:
By analyzing the blame information on this pull request, we identified @scalvert, @ry5n and @tedtate to be potential reviewers.

jvoll:
I am personally in favour of `var self = this` -- I find that I miss the .bind at the end of the block and that it is something I don't notice until I'm done reading the whole block at which point I have to reread/rethink the meaning of the block now knowing that the `bind` is present.

javascript/README.md
```diff
@@ -402,3 +402,48 @@ Prefer the short, more succinct version.
 Array.prototype.forEach.call(arguments, function(arg) { console.log(arg); });
 ```
 
+##Prefer `self` to `bind(this)`
+
```
jeremywiebe: pew pew

javascript/README.md
```diff
@@ -402,3 +402,48 @@ Prefer the short, more succinct version.
 Array.prototype.forEach.call(arguments, function(arg) { console.log(arg); });
 ```
 
+##Prefer `self` to `bind(this)`
```
jeremywiebe: Should we be more strong with the wording here?  "Always use" vs "Prefer".  Can anyone think of a reason to use `.bind()` where `self` wouldn't be just as easy/good?

jansepar:
This stack overflow answer has some interesting points:

http://codereview.stackexchange.com/questions/49872/using-var-self-this-or-bindthis

One that I thought was interesting - using `bind` takes away your ability to grab the original `this` that would have been in scope without the bind.

jansepar:
In general I think my vote would be :+1: to `var self = this`, unless anyone can sway me the other way :)

jansepar:
Also, I'd say neither approach is necessarily *wrong*, I would just like to standardize / bring some consistency to how we do it, because right now how we do it is dependant on the individual.

javascript/README.md
lizcross: I was just looking at some code I wrote. I wanted to pass in a prototype function into another function to be called at a later time. I think this is reasonable time to use `bind(this)`. Here is the line: https://github.com/mobify/astro-scaffold/blob/hyb-578-update-scaffold-header-bar-tabs/app/scaffold-controllers/navigationController.js#L38

DanielBrookRoberge:
I'm :+1: on this, there's some conceptual tidiness with `bind` but it's way too unobtrusive.

tedtate:
This is a bit of a side benefit of `self` over `bind` that I ran into recently when writing some tests:

```
var mySpy = sinon.spy(object, 'doThing');
$(window).on('event', object.doThing.bind(this));
```

If you write a test that uses spy/mock you won't be able to test the amount of times `doThing` is called due to it being a new method.

jansepar:
Thx for the feedback @tedtate :)

At this point I'm happy to :+1: - doesn't seem like there are any objections for using .bind by default.

@lizcross :ship: :shipit:

ericmuyser:
@tedtate Kind of curious, what if you did `object.doThing = object.doThing.bind(this)` before spy?

I know that would permanently change `object.doThing`, so you may want to unbind. I figure internally the interpreter is it's doing something similar to this anyway (capturing the scope to get 'self' and creating a closure). Thoughts?

tedtate:
@ericmuyser I probably should have written that out a little clearer:

```
class Obj {
    constructor():
        $(window).on('event', object.doThing.bind(this));
}
```

```
test():
    var object = new Obj();
    var mySpy = sinon.spy(object, 'doThing');
```

Spy creation happens after the fact so you don't really have that option.

ericmuyser:
Ah, that totally makes sense. Thanks @tedtate !

I've never seen anybody do this before, but now I'm wondering what the downside to [doing this would be](https://jsfiddle.net/jn14ya0r/1/)?

PR #85, opened 2016-01-13
[Objective-c] Added first version of objective-c language code-style description

AdamStreet:
Hey guys,

I started to work on my own keep-in-mind notes of the Objective-c language.

> This is not a full language-review with all best-practices, pitfalls or formatting sequences - just a very 
> rough set of the most common mistakes I have found in others' code yet.

Please take a view and feel free to comment - I accept any constructive criticism!

### DIFF
--- a/objective-c/README.md
+++ b/objective-c/README.md
@@ -1,0 +1,519 @@
+# Introduction
+
+This document is useful if you work on a huge project and want to keep it readable, clean & maintainable.
+Also helps you to avoid common mistakes (I have found since I started my career) and shows some best-practices.
+
+# Class naming convention
+
+Use class prefixes instead of just plain names. This makes your classes more unique - imagine if everybody would use plain names: you couldn't import a third party framework without conflict.
+> You can set up the default prefix in Xcode: select the project file in the Navigator (file browser, left side),
+> select File inspector tab and under Project Document / Class Prefix.
+
+# Markers (pragma mark)
+
+These markers can help your (and an other developer's) work to find methods.
+
+Recommended markers in order:
+
+* Initialization (-init, -dealloc, shared instance accessor of a singleton)
+
+* Private methods
+
+* View lifecycle (overrided methods of UIKit classes: -viewDidLoad, -viewWillAppear... etc.)
+
+* Public methods
+
+* Overrides
+
+* Accessors
+
+* Notification handlers
+
+* KVO handler
+
+* <[DelegateName]> (for delegate method implementations)
+
+Usage:
+`#pragma mark - [title]`
+
+or embed markers:
+
+`#pragma mark - [main title]`
+
+`#pragma mark [sub-title]`
+
+# Naming global constants
+
+Just like notification names, enum types... etc. They need to reflect to the container class, so 
+it should use it as a base. This help the user of this class to use code completion without checking the classes header.
+
+Example:
+
+	extern NSString * const MYClassDidSomethingNotification;
+
+	extern const CGFloat MYClassConstantFloatValue;
+
+	typedef NS_ENUM(NSInteger, MYClassSide)
+	{
+	MYClassSideLeft,
+	MYClassSidesRight
+	};
+
+	@interface MyClass : ...
+
+
+# Asserts
+
+Use asserts to keep thigt conditions in DEBUG mode. It will help you to find programming mistakes and/or
+invalid states in runtime.
+
+Examples:
+
+	- (BOOL)checkThis:(NSNumber *)number
+	{
+		NSParameterAssert(number);
+		...
+	}
+
+	- (void)myMethod
+	{
+		const NSUInteger numberOfItems = [self numberOfItems];
+		
+		NSAssert(0 < numberOfItems, @"No item has been found");
+		
+		...
+	}
+
+Handy, isn't it?
+
+> Avoid to use messages in the NSAssert body!
+> The asserts are included when you're in DEBUG mode - so the message won't be sent in RELEASE build 
+> which can change it's behaviour.
+
+> Note: NSParameterAssert became deprecated when the _\_Nonnull_ annotation was announced.
+> _- (void)checkThis:(NSNumber * \_Nonnull)number;_
+> This would help you when compiling the app instead of getting exceptions in runtime, 
+> but it's a real pain in the ass to add this to your existing codebase.
+
+# Initializers
+
+Always check the header for designated initializers! Override, or use it in your implementation.
+
+> If you create a new one and your class does't work normally with the
+> default initializer of it's superclass then mark it with `NS_DESIGNATED_INITIALIZER`
+
+Example:
+
+MYCar.h
+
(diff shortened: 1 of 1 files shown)


johnboxall:
By analyzing the blame information on this pull request, we identified  and undefined to be a potential reviewer.

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
+# Objective-c 
+
+## Introduction
+
+This document is useful if you work on a huge project and want to keep it readable, clean & maintainable.
+Also helps you to avoid common mistakes (I have found since I started my career) and shows some best-practices.
+
+## Class naming convention
+
+Use class prefixes instead of just plain names. This makes your classes more unique - imagine if everybody would use plain names: you couldn't import a third party framework without conflict.
```
jeremywiebe: App projects typically don't need class prefixes. You are in charge of all class names and removing the prefix makes for cleaner, easier to type names (`JJWCustomerView` vs `CustomerView`). Since you control the names of the classes in your project, should any class name conflicts arise with imported libraries, you can easily adjust your class name to resolve the conflict. In my experience collisions happen rarely and the easier to read/type class names are a real benefit.

For library projects using [3-letter](http://nshipster.com/namespacing/#3rd-party-class-prefixes) class prefixes is still a very good idea.

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+		const NSUInteger numberOfItems = [self numberOfItems];
+		
+		NSAssert(0 < numberOfItems, @"No item has been found");
+		
+		...
+	}
+
+Handy, isn't it?
+
+> Avoid to use messages in the NSAssert body!
```
jeremywiebe: Also don't depend on `NSAssert` to catch things at runtime in Release mode.  They're completely compiled out so if you hit that situation in a Release mode build, your program will continue on in a state you were trying to prevent.

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+
+* Use weak for storing delegates,
+
+* Copy for classes which have mutable versions (NSString, NSArray, NSDictionary...) and blocks,
+
+* Strong for everything else.
+
+### Local variable vs property
+
+Local variables are slowly being dropped from the framework - I recommend to declare properties all the time
```
jeremywiebe: Just to be clear, local variables are being declared by the compiler behind the scenes when you declare an `@property` in most cases.  It isn't if you define an `@property` but override the getter and setter.

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+* Account Manager
+
+* Uploaders and Downloaders... (most of the network layer classes)
+...
+
+> Note: Watch out when subclassing a singleton: the shared instance getter class method won't 
+> work in your subclass. You need to write the _+sharedInstance_ manually and 
+> keep your (static) _sharedInstance_ in your implementation
+
+If your singleton observes a property or reacts on a notification an initializer shoud be implemented which
```
jeremywiebe: `should`

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+
+		return self;
+	}
+
+	- (void)dealloc
+	{
+		[[NSNotificationCenter defaultCenter] removeObserver:self];
+	}
+
+> Use dedicated methods for handling a notification! Avoid to bind methods like _-clearTableView_.
```
jeremywiebe: Good point and good naming choice!

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+			options:(NSKeyValueObservingOptions)0
+			context:kvoContext];
+	}
+
+		return self;
+	}
+
+	- (void)dealloc
+	{
+		[otherObject removeObserver:self forKeyPath:@"parameter" context:kvoContext];
```
jeremywiebe: `dealloc` is not always a safe place to unregister. In some cases NSNotificationCenter will retain a reference to the object you register and so `dealloc` will never run until you unregister. I think this is mostly fixed in iOS 9, but like I said above, keep in mind on iOS 8 and lower.

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+
+	@property (nonatomic, readonly, strong) UIView *backgroundView;
+
+	@end
+
+MYViewController.m
+
+	@implementation MYViewController
+
+	@syntesize backgroundView = _backgroundView;
```
jeremywiebe: Since you recommend not declaring variables explicitly above this should probably be removed also, right?

objective-c/README.md
```diff
@@ -0,0 +1,522 @@
...
+		if (error) {
+			[weakSelf showErrorMessage:error];
+		} else {
+			[weakSelf updateContent:result];
+		}
+	}];
+
+## Some useful UIKit patterns
+
+### Properties of UIViewController
```
jeremywiebe: I think you could just name this "Lazy creation of expensive resources".  The pattern is useful all over the place where its expensive to create a resource and so you create it on demand. :)

jeremywiebe:
Hi @AdamStreet !  Thanks for this code style document. At Mobify, Objective-c isn't one of our used languages (we have a few corner projects that still have some Objective-c but we won't start any new projects with it).  As a result, we won't merge this PR, but I'd encourage you to fork this repo and apply these changes to your version. :)  You have some good recommendations and hopefully I've provided some useful info.

objective-c/README.md
AdamStreet: Exactly! 
There was a situation where an unexpected error caused a crash. An assert was set up in the code which had looked like this (silly example):

    NSAssert([self.fileManager removeItemAtPath:path error:nil], @"Unable to remove file");

Of course, it worked in DEBUG builds and just after a long code-review we found this line.

objective-c/README.md
AdamStreet: Well, depends on what we want to achieve with that read-only property.

If I needed a method to *create* a backgroundView I'd use a method like

    - (UIView *)buildBackgroundView {...}

This creates a new instance of UIView every time I send this message.

But in this case I want to get the `backgroundView` (which was already created) every time I ask for `self.backgroundView`, so I need to store it somewhere - and also need the `@syntesize myProperty;` because the [compiler doesn't do it for me in this case](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html#//apple_ref/doc/uid/TP40011210-CH5-SW13).

AdamStreet:
Hi @jeremywiebe !
Thank you for the feedback! I definitely will apply some changes based on your comments.
I understand your point - the Objective-c is slowly fading (or not that slowly) into the shadow of Swift, and we (iOS developers) must keep up.
So I'm preparing for my next code style document which is about Swift. :)

jeremywiebe:
Cool @AdamStreet.  You might want to see if https://github.com/mobify/mobify-code-style/tree/master/swift has anything useful.  :smile:

PR #86, opened 2016-01-26
Fix the point about function expression hoisting

aaani:
# Fix the point about function expression hoisting

Status: **Up For Review**

Reviewers: @jansepar 

## Changes
- In javascript, function expressions are not hoisted, but function declarations are. This doc said the exact opposite of that. I updated the doc with the correct info.

## How to test-drive this PR
- Read the section`Use function expressions over function declarations` and make sure it makes sense. 

## Research resources
- [Function declaration hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#Function_declaration_hoisting)


johnboxall:
By analyzing the blame information on this pull request, we identified @scalvert, @lizcross and @jansepar to be potential reviewers.

jansepar:
@aaani nice catch! :+1:

PR #89, opened 2016-02-01
Add reference to Swift API design guidelines.

jeremywiebe:
Adds a reference to the recently published Swift API design guidelines on swift.org. 

If we don't already follow these in all of our code, we definitely should. :)


johnboxall:
By analyzing the blame information on this pull request, we identified @dnerdy, @jeremywiebe and @jvoll to be potential reviewers.

kschmidtdev:
:+1:

PR #90, opened 2016-02-04
Add a stylelint ruleset

marlowpayne:
Status: **WIP**
Owner: @marlowpayne
Reviewers: @jeffkamo @mikenikles @jansepar 

## Changes
- SCSS-lint is the very last tool in our build chain that requires Ruby
- This change is in tandem with this generator PR: https://github.com/mobify/generator-adaptivejs/pull/343 , which migrates SCSS linting to [stylelint](http://stylelint.io/), finally breaking our dependency on Ruby
- This ruleset needs to be brought up to the feature parity with our existing [SCSS linting ruleset](https://github.com/mobify/mobify-code-style/blob/develop/css/.scss-lint.yml) to effectively enforce our SCSS code style

## Todos:
- [ ] Reach feature parity (within acceptable bounds) with our existing SCSS linting ruleset
- [ ] UI Dev +1

### Feedback:
_none so far_

## How to Test
- Check out this generator PR, which uses this ruleset: https://github.com/mobify/generator-adaptivejs/pull/343

PR #91, opened 2016-02-17
Add a sass-lint ruleset

marlowpayne:
Status: **Ready for Review**
Owner: @marlowpayne
Reviewers: @jeffkamo @mikenikles @jansepar 

## Changes
- Like #90, but even better because the feature diff seems to be pretty minimal
- This change is in tandem with this generator PR: https://github.com/mobify/generator-adaptivejs/pull/344 , which migrates SCSS linting to [sass-lint](https://github.com/sasstools/sass-lint), finally breaking our dependency on Ruby
- This ruleset needs reviewing/triaging against our existing SCSS linting ruleset to effectively enforce our SCSS code style
- This ruleset was generated using sass-lint's recommended tool for existing SCSS-lint users: http://sasstools.github.io/make-sass-lint-config/

## Todos:
- [x] Evaluate if this is a worth while change
- [x] UI Dev +1
- [ ] Cut a new release of code-style so that the generator no longer needs to point to this branch

### Feedback:
_none so far_

## How to Test
- Checkout the generator PR! https://github.com/mobify/generator-adaptivejs/pull/344


jeffkamo:
I've compared the Sass-Lint features against SCSS-Lint. See [that comparison here](https://docs.google.com/spreadsheets/d/1ZkMj_wzzPKvpMv7KjeECe3nkoP6qCqkDBKimTOL9KYg/edit#gid=761150116).

css/.sass-lint.yml
```diff
@@ -0,0 +1,350 @@
...
+# DisableLinterReason, ElsePlacement, PropertyCount, SelectorDepth
+# SpaceAroundOperator, TrailingWhitespace, UnnecessaryParentReference, VendorPrefixes
+# Compass::*
+#
+# The following settings/values are unsupported by sass-lint:
+# Linter Indentation, option "allow_non_nested_indentation"
+# Linter Indentation, option "character"
+# Linter SpaceBeforeBrace, option "allow_single_line_padding"
+# Linter PropertySortOrder, option "min_properties"
+# Linter PropertySortOrder, option "separate_groups"
```
jeffkamo: We can probably remove the above comments. Nice to know, but we have this information recorded elsewhere.

css/.sass-lint.yml
```diff
@@ -0,0 +1,350 @@
...
+        - animation-iteration-count
+        - animation-direction
+        - animation-fill-mode
+        - animation-play-state
+        - null
+        - cursor
+        - user-select
+        - resize
+        - pointer-events
+        - null
```
jeffkamo: I suspect these Null values will need to be removed. I'm fairly certain they don't add spaces between properties groups.

marlowpayne: Will do. sass-lint won't support separating groups anyways.

css/.sass-lint.yml
```diff
@@ -0,0 +1,350 @@
...
+  trailing-semicolon: 2
+  url-quotes: 2
+  variable-for-property:
+    - 2
+    - properties:
+        - font
+  variable-name-format:
+    - 2
+    - allow-leading-underscore: true
+      convention: '[a-z]|-{3}|_{3}|[^_]_[^_]'
```
jeffkamo: Have we tested that these three usages of regex work with our convention?

marlowpayne: @jeffkamo this regex is taken from our SCSS-lint `NameFormat` [rule](https://github.com/mobify/mobify-code-style/blob/develop/css/.scss-lint.yml#L132) which is actually lifted from the [SCSS-lint source code](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/selector_format.rb#L58). It's the same regex that SCSS-lint uses for hyphenated BEM format.

jeffkamo:
:+1:

mikenikles:
That looks good to me as well :+1:

PR #96, opened 2016-02-24
Adds a snippet for our recommended best practice around creating modules in JS

jansepar:
Adds a snippet for our recommended best practice around creating modules in JS

Status: **Opened for review**

Reviewers: Engineers @ Mobify

## Changes
- Defines our best practice around defining JS modules.

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -505,3 +505,32 @@ functionThatReturnsAPromise()
         // other code here!
     });
 ```
+
+## Module Definition
+
+When building a module that can be consumed by a number of different libraries, we prefer
+to use [Universal Module Definition (UMD)](https://github.com/umdjs/umd), so that the
+module will be compatible with AMD, CommonJS, or plain script inclusion.
+
+For example:
+
+```
+(function(root, factory) {
+    if (typeof define === 'function' && define.amd) {
+        // AMD. Register as an anonymous module.
+        define([], factory);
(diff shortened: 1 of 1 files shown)


javascript/README.md
```diff
@@ -505,3 +505,32 @@ functionThatReturnsAPromise()
         // other code here!
     });
 ```
+
+## Module Definition
+
+When building a module that can be consumed by a number of different sources, we prefer
```
jeremywiebe: What about modules that aren't consumed by a number of different sources... do we support a different module definition then?

jansepar: Nah - if it's just something your consuming that will only ever live in one project, it doesn't need to be UMD.

jansepar: I'll make that more clear.

javascript/README.md
```diff
@@ -505,3 +505,36 @@ functionThatReturnsAPromise()
...
+to use [Universal Module Definition (UMD)](https://github.com/umdjs/umd), so that the
+module will be compatible with AMD, CommonJS, or plain script inclusion.
+For modules that are not shared across projects, UMD is not required.
```
mikenikles: @jeremywiebe and @jansepar: Do you think that covers your discussion above? If so, let's get this merged.

jansepar: Yup looks good.

PR #98, opened 2016-03-01
[WIP] Update class naming conventions

jeffkamo:
Status: **WIP**
Owner: @jeffkamo
Reviewers: @cole-sanderson @bengroulx @mlworks @yourpalsonja @Magitron 
## Changes
- Start updating the Class Naming Conventions to match our modern best practices.
## Todos:
- [x] Developer +1 (given by @bengroulx)


css/class-naming-conventions/readme.md
```diff
@@ -20,14 +20,15 @@
 ## Basic Conventions
 
 * Class names are kebab-case (*words-are-dash-separated*)
-* Each class is prefixed with either `c-`, `t-` or `x-` ([consult this table](#class-prefix-conventions) for details)
+* Subclasses are indicated with double underscore, such as `root__subclass`
```
yourpalsonja: I'm not sure I understand what this `root__subclass` situation is.

css/class-naming-conventions/readme.md
```diff
@@ -20,14 +20,15 @@
...
+* Each class is prefixed with either `c-`, `t-` or `u-` ([consult this table](#class-prefix-conventions) for details and other rarer prefixes)
 
 
 ## CSM
 
-Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [BEM principles](http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/) to denote types of classes while still maintaining full use of the cascade.
+Our CSS class naming convention (which we call CSM or Component, Sub-Component, Modifier) uses the principles as popularized by [BEM](http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/) to denote types of classes while still maintaining full use of the cascade.
```
yourpalsonja: Nitpick, but this had less chunky-flowage before the edit. ("..uses BEM principles...")

css/class-naming-conventions/readme.md
```diff
@@ -43,23 +44,23 @@ Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [B
 
 This example may seem confusing at first but if we break down each of the selectors that we have, it begins to make more sense.
 
-`.c-blog` This is a component. It describes a high level module or component. In this instance, it describes the container for all of our blog posts.
+`c-blog` This is a component. It describes a high level module or component. In this instance, it describes the container for all of our blog posts.
```
yourpalsonja: "This is a high-level component" might be enough here. What do you think about avoiding the word "module" so that our language stays consistent? I know components were sometimes called modules in old projects, but we don't do that anymore. ✂️

css/class-naming-conventions/readme.md
```diff
@@ -43,23 +44,23 @@ Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [B
...
 
-`.c-blog__title` This is a sub-component. It's always a child of a module or component. In this instance, it is a title for our blog post container
+`c-blog__title` This is a sub-component. It's always a child of a module or component. In this instance, it is a title for our blog post container
 
-`.c-blog-post` This is another component. This one describes a specific blog post. We make this its own component because a blog post is not necessarily a child of the blog container. It can and should be able to live independently.
+`c-blog-post` This is another component. This one describes a specific blog post. We make this its own component because a blog post is not necessarily a child of the blog container. It can and should be able to live independently.
```
yourpalsonja: Is it true that `c-blog-post` might not always be a child of `c-blog`? I was under the impression that sub-components always had to live inside the component itself–unless what we're trying to say here is that it doesn't have to be a direct child of `c-blog`.

yourpalsonja: Oops, rescinding my previous comment, I read that as `c-blog__post`!

css/class-naming-conventions/readme.md
```diff
@@ -68,14 +69,13 @@ The highest level of a module — it should describe an independent module that
 
 ### Sub-components
 
-This is a secondary element inside of a component. It is always written as a chain of its parent component to avoid any inheritance issues. Your subcomponents should be named in a way that keeps them from having to have subcomponents of their own. If you find you need to write a subcomponent for a subcomponent, consider breaking the parent out into its own component.
+This is a secondary element that is child to its parent component. The classname should be formatted as such: `c-[parent-component-name]__[sub-component-name]`. Sub-components do not, and should not, have sub-components of their own. If you find you need to write a sub-sub-component, instead just treat it as a sub-component – sub-components are only ever child to the parent component.
```
yourpalsonja: "A sub-component is a child of its parent component, with a classname like `c-[parent-component-name]__[sub-component-name]."

Should we also mention that if you need to write a bunch of sub-sub-components, you might be making your component too "high-level"?

css/class-naming-conventions/readme.md
```diff
@@ -68,14 +69,13 @@ The highest level of a module — it should describe an independent module that
...
 
-Like components these should always live at the root level of a file. Do not nest these within the parent component or another subcomponent. The class name should do all the work necessary.
+Like components these should always live at the root level of a file. Avoid nesting these within the parent component or another sub-component.
```
yourpalsonja: Missing a comma here after "Like components".

Should we make it clear that they live at the root level of a CSS file? (Just to be explicit). Or maybe we could frame this more as "sub-components shouldn't be styled as descendants of their parent component (e.g. no `.c-component .c-subcomponent` stuff going on inside the CSS file).

css/class-naming-conventions/readme.md
```diff
@@ -89,15 +89,23 @@ Like components these should always live at the root level of a file. Do not nes
...
 ### Modifiers
 
-These are used to modify components or subcomponents. They are always chained to a specific component and are declared in the component or subcomponent that they affect.
+Modifiers, as their name suggests, modify components or sub-components. They are always chained to the component or sub-component they belong to.
 
 * Prefixed with the namespace of the affected element and two dashes (`c--`, `t--`)
 * Contained to the scope of a single component
-* Always declared as a chained selector to a component or subcomponent.
+* Always declared as a chained selector to a component or sub-component.
 * Never declared as a stand-alone rule.
```
yourpalsonja: Should we link to the *Modifiers that Affect Sub-components* section of the doc here?

css/class-naming-conventions/readme.md
```diff
@@ -264,39 +279,45 @@ An exception is the use of ARIA roles for styling state. Where an ARIA role maps
...
+* When you find yourself using @extend
 
 ### When to use their existing selectors
 
-* Whenever possible — when you're not required to do any of the things above. It's faster and easier to use their markup than it is to add our own.
-* When their markup allows for it. For example, if they don't use classes or they don't use them with any consistency, it doesn't make sense to use their selectors.
-* When their markup isn't easily changed. AJAXed content or content added after a page is loaded is an example of this.
+* When it's fastest, easiest or most efficient to use their markup than it is to add our own.
+* When their markup is too inconsistent, or makes parsing too difficult.
+* When desktop functionality is tightly coupled to desktop's markup structure.
```
yourpalsonja: Is this meant to say "when *mobile* functionality..."

css/class-naming-conventions/readme.md
```diff
@@ -43,23 +44,23 @@ Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [B
 
 This example may seem confusing at first but if we break down each of the selectors that we have, it begins to make more sense.
 
-`.c-blog` This is a component. It describes a high level module or component. In this instance, it describes the container for all of our blog posts.
+`c-blog` This is a component. It describes a high level module or component. In this instance, it describes the container for all of our blog posts.
 
-`.c-blog__title` This is a sub-component. It's always a child of a module or component. In this instance, it is a title for our blog post container
+`c-blog__title` This is a sub-component. It's always a child of a module or component. In this instance, it is a title for our blog post container
```
yourpalsonja: Should we avoid the word `module`?

Also, missing a period here. :)

css/class-naming-conventions/readme.md
```diff
@@ -43,23 +44,23 @@ Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [B
...
 
-`.c-blog-post` This is another component. This one describes a specific blog post. We make this its own component because a blog post is not necessarily a child of the blog container. It can and should be able to live independently.
+`c-blog-post` This is another component. This one describes a specific blog post. We make this its own component because a blog post is not necessarily a child of the blog container. It can and should be able to live independently.
 
-`.c--featured` This is a modifier. It is always chained to a component or sub-component. In this instance, it describes a different way of displaying a component.
+`c--featured` This is a modifier. It is always chained to a component or sub-component. In this instance, it describes a different way of displaying the `c-blog-post` component.
 
-`.c-blog-post__time` Like before, this is another sub-component. This time it belongs to the c-blog-post. It's still a subcomponent even though it is not a direct child of the component.
+`c-blog-post__time` Like before, this is another sub-component. This time it belongs to the c-blog-post. It's still a sub-component even though it is not a direct child of the component.
```
yourpalsonja: "A sub-component belonging to `c-blog-post`. Sub-components aren't required to be direct children of their parent component."

css/class-naming-conventions/readme.md
```diff
@@ -68,14 +69,13 @@ The highest level of a module — it should describe an independent module that
...
-Like components these should always live at the root level of a file. Do not nest these within the parent component or another subcomponent. The class name should do all the work necessary.
+Like components these should always live at the root level of a file. Avoid nesting these within the parent component or another sub-component.
 
-* Prefixed by the parent component and two underscores `c-component-name__`.
-* Live below the parent component in the root of the file. Not nested.
-* Are declared in the order they appear.
-* Subcomponents do not have to be direct children of the component in the markup. They can be any descendent.
+* Prefixed by the parent component and two underscores `c-[component-name]__[sub-component-name]`
+* Lives below the parent component in the root of the file, un-nested
+* Subcomponents do not have to be direct children of the component in the markup. They can be any descendent
```
yourpalsonja: Can probably remove "they can be any descendent"...

css/class-naming-conventions/readme.md
```diff
@@ -264,39 +279,45 @@ An exception is the use of ARIA roles for styling state. Where an ARIA role maps
...
+`.m-` (*) | Desktop embedded mobile markup classes: these are classes that we will use if we author Markup that is intended for clients to embed onto their desktop pages, but is for mobile content. | *n/a* |
 `.js-` | Javascript classes are used exclusively by scripts and should never have CSS styles applied to them. Repeat: **Do NOT** style Javascript classes. | *n/a*
 
+> * The `m-` class prefix has an old, deprecated use: Mobify Modules. However, Mobify Modules have been replaced with plugins, and are treated as third party libraries with their own conventions.
+
+
+## Parsing vs. Decorating
 
-## Us versus Them (aka There's an x-ception to every rule)
+It's important to understand that we have a few different ways of authoring our CSS, and the way we do this is depends a lot on how we convert the desktop markup for mobile. On one hand, we parse the desktop markup and take full control of the final HTML. On the other hand, sometimes we just output the desktop markup as is, untouched or perhaps wrapping certain chunks, and instead control the appearance entirely through CSS.
```
yourpalsonja: I think we need to put an emphasis on the fact that using CSM is the preferred option, and that  having selectors like `.c-blog-post p` or `.c-blog-post span` should be avoided if possible.

Perhaps rather than saying "On one hand..." and "On the other...", we could say "Ideally..." and "However, this isn't always possible, and..."

css/class-naming-conventions/readme.md
```diff
@@ -264,39 +279,45 @@ An exception is the use of ARIA roles for styling state. Where an ARIA role maps
...
 
-It's important to remember that we don't write our own markup. We write a bastardized version of existing markup. In many cases, we're simply adding wrappers or class names to markup that already exists. Rarely, we'll completely re-template something.
+If you find yourself wondering "should I be adding a new class, or should I use the classes from desktop?" consider the following: If we're using their class names, we obviously can't follow our CSM syntax. But that said, sometimes we just have no choice; perhaps there are engineering requirements that force us to retain the markup structure. Under such circumstances, we must stick to the desktop classes.
```
yourpalsonja: I think we could make it even clearer here that CSM is always the best option. Maybe these two paragraphs could be combined into one.

css/class-naming-conventions/readme.md
```diff
@@ -43,23 +44,23 @@ Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [B
...
+`c--featured` This is a modifier. It is always chained to a component or sub-component. In this instance, it describes a different way of displaying the `c-blog-post` component.
 
-`.c-blog-post__time` Like before, this is another sub-component. This time it belongs to the c-blog-post. It's still a subcomponent even though it is not a direct child of the component.
+`c-blog-post__time` Like before, this is another sub-component. This time it belongs to the c-blog-post. It's still a sub-component even though it is not a direct child of the component.
 
 ### Components
 
-The highest level of a module — it should describe an independent module that you are creating. Components should be able to exist on their own or within other components. They should always live at the root level of a file.
+The highest level of a module — it should describe an independent, self-contained module. Components should be able to exist on their own or within other components. They are only responsible for itself or for what is within itself, never for anything external to itself. Component classes live at the root level of a file.
```
yourpalsonja: Components are independent and self-contained. Styles belonging to a component should only affect the component itself, and any of its sub-components. They shouldn't affect anything external to them, or any other components that might be nested within them. (?)

Or change 'itself' to 'themselves'. :)

css/class-naming-conventions/readme.md
jeffkamo: I'll clarify in the example code a little lower down (where the good/bad examples are)

css/class-naming-conventions/readme.md
jeffkamo: I'm not really sure it's necessary. It's the very next section, making it not very far away.

css/class-naming-conventions/readme.md
jeffkamo: Either that, or rewrite the paragraph to offer more concrete answers to the questions it raises.

yourpalsonja:
Added @Magitron as a reviewer!

css/class-naming-conventions/readme.md
yourpalsonja: Fair enough.

yourpalsonja:
Hey, I'm not sure what part of the doc this can fit into, but I've also run across problems recently where Component A is nested inside Component B, and styled by it as well. 

Contrived example: something that would compile as: `.c-button .c-icon`, where `.c-icon` is styled when it is within a button.

In this case, in my opinion, you can give the icon both the `c-icon` and `c-button__icon` classes, so that there is a hook to style that belongs exclusively to `c-button` and doesn't require increased specificity when it needs to be slightly tweaked (either by a modifier or a template class). 

Thoughts?

jeffkamo:
@yourpalsonja good point! I've added a new section that discussed that point. I am in favour of adding sub-component classes to external components to prevent you from styling things a component shouldn't be aware of.

jeffkamo:
I think that was the last thing I needed to address. Final thoughts, anyone?

bengroulx:
Is it worth tossing in the `pw-` prefix somewhere in here?

css/class-naming-conventions/readme.md
```diff
@@ -43,23 +45,23 @@ Our convention (which we call CSM or Component, Sub-Component, Modifier) uses [B
...
 
-`.c-blog-post__time` Like before, this is another sub-component. This time it belongs to the c-blog-post. It's still a subcomponent even though it is not a direct child of the component.
+`.c-blog-post__time` Like before, this is another sub-component this time belonging to the `c-blog-post` component. It is still a sub-component even though it is not a **direct** child of the component. In other words, sub-components are not required to be direct children of its parent component.
 
 ### Components
 
-The highest level of a module — it should describe an independent module that you are creating. Components should be able to exist on their own or within other components. They should always live at the root level of a file.
+Components are independent and self-contained units of UI. Styles belonging to a component should only affect the component itself, and any of its sub-components. They should not affect anything external to them, or any other components that might be nested within them.
```
bengroulx: It’s not explicitly correct, but often component styles are coupled to a specific set of markup. May be worth mentioning.

jeffkamo: I'm not sure I agree with that. For 99% of the time, the markup shouldn't technically matter. When it does (i.e. when we have to target the anchor version of a button to apply `inline-table`) those are the exception and not the rule.

Is that what you are referring to?

yourpalsonja:
@jeffkamo can this be merged? 👀

jeffkamo:
Ya, let's merge it now. I foresee more changes and updates coming in the future, so we can review this more at that time.

PR #104, opened 2016-03-26
Remove obsolete options

amovah:


benlast:
The changes are appreciated, but right now we don't have a need to change this config.

PR #108, opened 2016-07-06
ES6 draft standard

DanielBrookRoberge:
This PR adds a new-from-the-ground-up style guide and lint config for ES6 React builds. 

## TODOs:
- [x] +1
- [x] Updated README
- [x] Updated CHANGELOG


es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+- `eslint-plugin-import`
+- `eslint-import-resolver-webpack`
+
+The lint configuration is the definitive source for rules; this document explains the most notable ones. `eslint <source-files> --fix` will fix most formatting issues, such as spacing. 
+
+## Overriding Lint Rules
+
+Some of the lint rules disallow uncommon but valid behaviour that is easily confused with/typoed from much more common behaviour. If you need to use the disallowed behaviour on purpose, use an explicit lint override in the source.
+
+For example, the `no-bitwise` rule disallows the bitwise operators `&` and `|` due to their unpredictable (but valid!) behaviour when substituted for the logical operators `&&` and `|`. To use bitwise operations in a situation that's approprate, do the following:
```
MikeKlemarewski: Did you mean "`&&` and `||`" in the second occurrence?

es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+```javascript
+// Mask off the bottom four bits of the octet
+/* eslint-disable no-bitwise */
+let maskedValue = rawValue & 0xf0
+/* eslint-enable no-bitwise */
+```
+
+## Removing Lint Rules
+
+These rules are extensive and a work in progress. If you find any of the rules to be a hindrance in your own work, you can disable them in the `.eslintrc.yml` file.
```
MikeKlemarewski: I would encourage people to open a PR against these rules if they think they are a hindrance rather than to just override them. Or open an issue, then we can discuss them further.

es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+/* eslint-enable no-bitwise */
+```
+
+## Removing Lint Rules
+
+These rules are extensive and a work in progress. If you find any of the rules to be a hindrance in your own work, you can disable them in the `.eslintrc.yml` file. 
+
+# Base Language
+
+## Always use curly brackets with control statements
```
MikeKlemarewski: Is this specific to ES6? I feel like this may be a rule that we use for ES5. We can probably treat this README as an extension of the ES5 one. Document any new rules, or rules that override ES5 ones.

MikeKlemarewski: https://github.com/mobify/mobify-code-style/tree/develop/javascript#use-curly-braces-even-if-the-block-only-has-one-line

MikeKlemarewski: A counter argument may be: we're going to be writing ES6 from here on out, so it may be better to have more rules in here.

I may have just convinced myself that this belongs here :)

es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+const func2 = () => { console.log(2) }
+const func3 = () => { console.log(3) }
+
+module.exports = {
+    func1,
+    func2,
+    func3
+}
+
+// Good
```
MikeKlemarewski: Think we want to show the `good` condition? Or just the `better` one

es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+
+Formatting output with string concatenation is verbose and requires care to space correctly. With an ES6 template string, this becomes much clearer.
+
+```javascript
+// Bad
+return 'We saw ' + count + ' errors in the ' + category + ' category'
+
+// Good
+
+return `We saw ${count} errors in the ${category} category'
```
MikeKlemarewski: The ending `'` should be a ```

es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+	let sum = 0
+	for (let i = 0; i < arguments.length; i++) {
+		sum += arguments[i]
+	}
+	return sum
+}
+
+// Good
+// values is an Array so it has all the array methods
+const mySum = (...values) => {
```
MikeKlemarewski: This example is a bit confusing to me since a `mySum` method would probably only take one argument, the array to sum, so it might not be the best example for a rest parameter

MikeKlemarewski: Ahhh, I see, you can pass a variable number of arguments, and this grabs them as an array called `values`

es6/README.md
```diff
@@ -0,0 +1,406 @@
...
+// Good
+const MyComponent = (props) => {
+	return (
+		<ul> 
+			{props.items.map((item) => <li>{item.content}</li>}
+		</ul>
+	)
+}
+
+// Better
```
MikeKlemarewski: Again, I think we should just show the `better` case here.

MikeKlemarewski:
🎊  this is awesome 🎊

noahadams:
do we want to do separate es6 recommendations and layer jsx recommendations/linter rules on top?

I can think of a few places where we might want to adopt es6 (build tooling, other node scripts, Astro JS, duncan) and not use react/jsx...

DanielBrookRoberge:
@noahadams I'd forgotten I was going to do that. (the additional rules won't interfere for code with no JSX, but they will require installing the React plugin for eslint).

es6/README.md
```diff
@@ -0,0 +1,403 @@
+# Linting
+
+Included is a default ESLint configuration, `mobify-es6.yml` written for ESLint 3.x. It requires the additional modules:
+
+- `eslint-plugin-import`
+- `eslint-import-resolver-webpack`
```
MikeKlemarewski: Can we have these be dependencies of `mobify-code-style` so we don't have to add them to the projects that want es6 linting?

DanielBrookRoberge: We could do that, though it would also install them for ES5 projects that use the `mobify-code-style` module.

README.md
```diff
@@ -29,6 +29,23 @@ eslint:{
...
+```yaml
+extends:
+  - 'node_modules/mobify-code-style/es6/mobify-es6.yml'
+```  
+
+and for a React/JSX project:
+
+```yaml
+extends:
+  - 'node_modules/mobify-code-style/es6/mobify-es6-react.yml'
```
MikeKlemarewski: For these, I had to add a `./` to the front of the path. So: `./node_modules/mobi.....`

es6/README.md
MikeKlemarewski: Yeah, I guess minimizing unnecessary npm module installs sounds better in this case.

Do we also need `eslint-plugin-react`? That was included in the scaffold PR

jansepar:
Should we move the ES5 stuff into an ES5 folder like we did with ES6? Or maybe that will break backwards compatibility...?

es6/README.md
```diff
@@ -0,0 +1,403 @@
...
+// Bad 
+console.log(myObj['prop'])
+
+// Good
+console.log(myObj.prop)
+``` 
+
+## Use ES6 `import` rather than `require`
+
+ES6 provides a module syntax that we can use rather than a `require` function. Babel's implementation of the ES6 module loader uses CommonJS under the hood, so we can `import` external CommonJS modules with no problem.
```
jansepar: I believe we can also import RequireJS modules without a problem, but it's not clear to me if it's babel or webpack that enables that...

DanielBrookRoberge:
Yeah, if anyone is using the instructions in the README to run ES5 linting moving it to a different folder name will break their linting setup.

es6/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+```javascript
+// Bad
+const Moment = require('moment')
+const MyModule = require('./my-module')
+
+// Good
+
+import Moment from 'moment'
+import * as MyModule from './my-module'
```
jansepar: I feel like we should make it "Bad" to import *. It's especially bad for web since we want to be conscientious about file size...

es6/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+export const func1 = () => { console.log(1) }
+export const func2 = () => { console.log(2) }
+export const func3 = () => { console.log(3) }
+```
+
+
+## Don't use semicolons
+
+With ES6 (and the current lint rules!) we have finally arrived at an environment where automatic semicolon insertion won't cause problems. So don't use them.
```
jansepar: 😱

es6/README.md
MikeKlemarewski: I don't think so. Sometimes you want everything. Like in a reducer, you may import every action and have a reducer for each action.

es6/README.md
MikeKlemarewski: I see, the react plugin is in the root readme

MikeKlemarewski:
@DanielBrookRoberge I'm a 👍  on this. Would be good to get it merged so we can import it into the scaffold PR

es6/README.md
jansepar: Ah fair enough!

PR #110, opened 2016-07-26
Update PULL_REQUEST_TEMPLATE.md

jansepar:
## Changes
- Adds changes based on @johnboxall's Code Review talk

### DIFF
--- a/github-templates/PULL_REQUEST_TEMPLATE.md
+++ b/github-templates/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,9 @@ Linked PRs: (links to corresponding PRs, optional)
 ## TODOs:
 - [ ] +1
 - [ ] Passes tests
+- [ ] "How to test" steps have been run through
 - [ ] Updated README
 - [ ] Updated CHANGELOG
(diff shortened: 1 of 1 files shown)


github-templates/PULL_REQUEST_TEMPLATE.md
```diff
@@ -15,7 +15,9 @@ Linked PRs: (links to corresponding PRs, optional)
 ## TODOs:
 - [ ] +1
 - [ ] Passes tests
+- [ ] "How to test" steps have been run through
```
kschmidtdev: Might be implicitly understood here, but we would prefer that the testing was also successful :)

jeremywiebe: I actually feel like this is a checklist for the author, not the reviewer.... so does this make sense here?  The assumption (as Karl points out) is that the reviewer uses the How to Test section and tests.

jansepar: Ah, good point. Yeah the +1 essentially means that the reviewer has ran those steps. I'll remove.

marlowpayne:
👍 :shipit:

PR #111, opened 2016-07-27
Remove length restriction & strict rule

MikeKlemarewski:
Linked PRs: (links to corresponding PRs, optional)

## Changes
- Remove line length restriction. This turned out to be a pain almost immediately. Long lines should be caught in code review if they go beyond your laptop screen width.
- Remove the `strict` rule as sometimes we need 'use strict' in modules that are run through node.

## How To Test
- Checkout this branch and run `npm link`
- In a project that includes mobify-code style: run `npm link mobify-code-style`
- Write a line that is longer than 100 characters and run the lint rules, make sure they pass
- Add a 'use strict' declaration to the top of a module, run the lint rules and make sure they pass

## TODOs:
- [ ] +1
- [ ] Updated README
- [ ] Updated CHANGELOG

### DIFF
--- a/es6/mobify-es6.yml
+++ b/es6/mobify-es6.yml
@@ -141,9 +141,6 @@ rules:
   max-depth:
     - error
     - max: 5
-  max-len:
-    - error
-    - code: 100
   max-nested-callbacks:
     - error
     - max: 3
@@ -188,7 +185,6 @@ rules:
   space-infix-ops: error
   space-unary-ops: error
   spaced-comment: error
-  strict: error
   template-curly-spacing: error
   unicode-bom:
     - error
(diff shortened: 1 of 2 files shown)


es6/mobify-es6.yml
```diff
@@ -188,7 +185,6 @@ rules:
   space-infix-ops: error
   space-unary-ops: error
   spaced-comment: error
-  strict: error
```
fractaltheory: Yeah see I don't think this is really necessary to remove? When I think of the vast majority of code we'll apply linting to (i.e. projects), we'll want to enforce that `use strict` is unnecessary.

MikeKlemarewski: IMHO I rarely see 'use strict' in any files at Mobify, and even though it's unnecessary I don't think it's harmful to have it in a file where it's not needed. I'd rather let a few unnecessary 'use strict' definitions go through than encourage disabling linting rules.

fractaltheory:
👍

PR #128, opened 2016-08-26
✨ Add a git standards doc

marlowpayne:
## Changes
- 📚 Add missing entries to CHANGELOG
- ✨ Add a git standards doc
- 📚 Update CHANGELOG and README
## How To Test
- Read the new doc
- Discuss your thoughts
- Relevant portion of [Atom's CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#git-commit-messages)
- Relevant portion of [Sass-lint's CONTRIBUTING guide](https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md#git-commit-messages)
## TODOs:
- [ ] +1
- [x] Updated README
- [x] Updated CHANGELOG


git/README.md
```diff
@@ -0,0 +1,45 @@
...
+Git is pretty awesome. Here's some reasonable standards when it comes to using it.
+
+## Commit Messages
+
+At Mobify, we recommend a few standards regarding git commit messages:
+- If the commit is related to a JIRA ticket, prefix the case-sensitive ticket number to the start of the message
+  - The JIRA/GitHub integration is case-sensitive
+- Next choose a [relevant emoji](#emoji-list) to categorize your commit
+- Finally, your main message content should follow the [git convention](http://git.kernel.org/cgit/git/git.git/tree/Documentation/SubmittingPatches#n117) of present-tense, imperative mood
+  - "...as if you are giving orders to the codebase to change
```
jeremywiebe: Nitpick, but this section has a mix of hard line breaks and no line breaks...

yourpalsonja:
I whole-heartedly support this, +1, will add gif later.

git/README.md
marlowpayne: Thanks @jeremywiebe ! Not really sure how I managed that... 😕

git/README.md
```diff
@@ -0,0 +1,44 @@
+# Git
+
+Git is pretty awesome. Here's some reasonable standards when it comes to using it.
+
+## Commit Messages
+
+At Mobify, we recommend a few standards regarding git commit messages:
+- If the commit is related to a JIRA ticket, prefix the case-sensitive ticket number to the start of the message
```
jeremywiebe: Features or bug fixes that may span multiple commits might make this cumbersome.  I think our approach is to have the PR reference the JIRA ticket in the title "HYB-132 - Fix header bar offset" and also link in the description.  Do you think having every commit prefixed is needed?

marlowpayne: A couple thoughts:
- JIRA + GitHub integration requires that git commit messages start with a case-sensitive JIRA ticket number in order to work.
- Not all commits have adequate comments in their code changes. The JIRA ticket in the commit message helps to build the paper trail for getting context to a particular change.

jeremywiebe: * Not all projects have JIRA + GitHub integration. So forcing the case for that reason doesn't make sense. 
* Personally, I think those details should be in the PR and not in the commit message. Commits can be super small, even breaking changes, to a project. The PR is the atomic change. For any given commit you're only ever one click away from the PR that introduced it.

yourpalsonja: Is it possible to trace back a commit to the PR that it belonged to? On CS projects (especially older ones), I find it difficult to find the reason for a change unless the ticket # is mentioned in the git commit message (or a comment in the file, which is rare). Having the ticket # right there can help you find the JIRA ticket/PR easily.

jeremywiebe: @yourpalsonja Yes.  Each commit that was merged as part of a PR has a link back to the PR on it. 

screen_shot_2016-08-26_at_1_53_59_pm

marlowpayne: @jeremywiebe a few additional points I've been thinking of:
- While tracing a code change through commit to PR to JIRA ticket is possible, JIRA ticket numbers in commit messages mean that you have a double-linked paper trail: you can find commits based on a given JIRA ticket.
- While PRs are our new standard of atomic code operations, it means that relying solely on them for uncovering history and context is being locked into GitHub's interface. With JIRA ticket numbers in commit messages, you can search for them in _any_ git client.
- Discipline in commit messages is exactly what I'm looking for with these standards. Getting developers to micro-document their changes through commit messages means less work for maintainers down the line.

mikenikles: I'm a +1 for structured commit messages. In fact, for the Progressive Web SDK, we are moving to a very strict format for commit messages:

```
<type>(<scope>): <subject>
<BLANK LINE>
<body>
<BLANK LINE>
<footer>
```

Based on that, we will auto-generate the `CHANGELOG.md` file as well as automatically create a release and publish it to NPM and Github. This is needed in order to move closer to continuous integration.

Regardless of this upcoming change in our release process, I believe we're better off not relying on the Jira <> Github integration. Most often when it comes to maintenance, I open Github and check a files history to figure out what happened.

yourpalsonja:
@marlowpayne What's up with this? Should it be something that gets discussed at Web Dev Sync-up, or Engi meeting?

marlowpayne:
@yourpalsonja yesss. I'd like to see this idea get some more visibility and discussion again. I feel like highly structured git commits like @mikenikles alluded to are too large of a diff, and potentially a step too far. I would like to see us reach an agreement of our minimal git standards, ship that for internal devs and partners, and iterate from there.

jeremywiebe:
@yourpalsonja @marlowpayne Given that it impacts all of engineering, I'd say let's discuss at Engi meeting rather than Web Dev Sync-up. :)

jeffkamo:
I've asked to have this added to the Engineering meeting agenda!

git/README.md
benlast: On Pusheen we used the JIRA ticket number in the branch name. Simple, and necessary since we often have multiple commits for a single ticket as we do refactoring in stages.

git/README.md
```diff
@@ -0,0 +1,45 @@
...
+  - The JIRA/GitHub integration is case-sensitive
+- Next choose a [relevant emoji](#emoji-list) to categorize your commit
```
benlast: Can we at least make this optional? 🙄

git/README.md
sap1ens: @benlast I think that is totally enough for visibility, but not enough for tooling (like JIRA integration). Definitely great step forward.

marlowpayne: We do have a section in our [branching strategy docs](https://github.com/mobify/branching-strategy/blob/master/README.md#branch-naming) re. branch naming after JIRA tickets.

jeremywiebe: We have quite a few different viewpoints on this point:
* Prefix every commit with JIRA number (@m)
* Super-structured commits (@mikenikles)
* Branch names prefixed with JIRA number (@benlast)
* PR title or description includes JIRA number (me)

I think each has a different driver based on the workflow and needs of the person suggesting it. Is it worth laying down a standard that may not fit each team it applies to? 

@jvoll also brought up a good point: what about partners? Do we expect them to follow this standard?

jvoll: I don't think requiring a JIRA ticket number in each commit message fits inside the workflow of many of the product teams particularly. I do understand why traceability is a huge asset for figuring out why change X was made, especially from a maintenance perspective.

I think it's pretty reasonable to ask for branches + PRs to have a JIRA ticket number in them. But even that, sometimes I create branches that don't have a JIRA ticket. This can't be a rule, but maybe a guideline. An example, if I'm going to fix 2-3 typos in some docs on astro.mobify.com, I'm not going to create a JIRA ticket for that. As a general rule on our team we aren't creating JIRA tickets for things that come up that take less than ~30 minutes since creating a ticket actually adds 2+ minutes to this which is a non-trivial percentage of the actual work involved.

IMO -- we shouldn't need to change our process to fit the tool. If Github + JIRA integration requires changing commit messages in a way that is going to slow down the entire team, we should modify the tool, not our process.

RE partners: We already have enough to train partners on and even with our better partners we've had a bit of an uphill battle (in my experience) to get them to open PRs at the correct time and use good branching strategy. If partners are building all of our projects then these rules really apply to them more than Mobify-internal. I'm concerned that they have enough to learn already with the Mobify platform that this might not be a battle worth fighting just yet.

git/README.md
```diff
@@ -0,0 +1,45 @@
...
+- Finally, your main message content should follow the [git convention](http://git.kernel.org/cgit/git/git.git/tree/Documentation/SubmittingPatches#n117) of present-tense, imperative mood
+  - "...as if you are giving orders to the codebase to change its behaviour."
+
+E.g. `ABC-123: 🐛 Make feature X work by doing Y`
+
+## Emoji List
+
+Commit Type            | Emoji
+---                    | ---
+App/Astro              | 🚀 `:rocket:`
```
jvoll: not to hate on emojis... but it probably doesn't make sense to have a 🚀  emoji on every single astro commit. I know it's "optional" but I'm trying to think when using this would actually be a good idea.

yourpalsonja: Maybe emoji can be optional? I find them helpful for picking out a particular commit from a large list of them. If every commit message had a 🚀 , it wouldn't be very helpful. I would say the 🚀  might be more for a FEE/UI Dev making a change for Astro (in terms of styling or tooling), not the something the Astro team would use on each commit.

jansepar:
Hey folks, finally getting to this, sorry for the late reply.

Not opposed to having some standards on Git but I'm not seeing the value that this standards PR is going to provide. To me, the idea of having to use the right emoji for the right commit sounds like it's going to bring a lot of headache with not much upside. Even though it's stated as optional, I don't know that I even want to make it an optional recommendation because I'm not seeing the benefits.

For me, if we have any standards, they would be:

- If there is a JIRA ticket, prefix your branch with it.
- Make sure that your commit accurately and succinctly describe the change.

Thoughts?

jansepar:
Going to close this for now - if someone wants to push it forward feel free to re-open or open a new PR!

PR #129, opened 2016-08-29
Add jsx-a11y.

mikenikles:
Adds accessibility rules for React projects.

## Changes
- Add `eslint-plugin-jsx-a11y`

## How To Test
- Use this branch in a React project
- Add the following image component in one of the `render()` functions: ``
- ESLint should warn you that the image component needs a `alt` prop.

- https://github.com/evcohen/eslint-plugin-jsx-a11y

## TODOs:
- [ ] +1
- [x] Updated README
- [x] Updated CHANGELOG
- [x] Move a11y lint rules into it's own file

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 Pending Release
     - Remove `variable-for-property` rule from sass-lint
+    - Add the jsx-a11y plugin to the mobify-es6-react configuration.
 2.5.3
     - Fix #123 by using react/jsx-wrap-multilines rule
     - Migrate CSSComb documentation from Confluence to code style repo
--- a/es6/README.md
+++ b/es6/README.md
@@ -4,6 +4,7 @@ Included is a default ESLint configuration, `mobify-es6.yml` written for ESLint 
 
 - `eslint-plugin-import`
 - `eslint-import-resolver-webpack`
+- `eslint-plugin-jsx-a11y`
 
 The lint configuration is the definitive source for rules; this document explains the most notable ones. `eslint <source-files> --fix` will fix most formatting issues, such as spacing.
 
--- a/es6/README.md
+++ b/es6/README.md
@@ -4,11 +4,15 @@ Included is a default ESLint configuration, `mobify-es6.yml` written for ESLint 
 
 - `eslint-plugin-import`
 - `eslint-import-resolver-webpack`
-- `eslint-plugin-jsx-a11y`
 
 The lint configuration is the definitive source for rules; this document explains the most notable ones. `eslint <source-files> --fix` will fix most formatting issues, such as spacing.
 
 If React/JSX is in use, use the React default configuration `mobify-es6-react.yml`, which also requires the `eslint-plugin-react` module.
+
+JSX accessibility linting is available with the `mobify-es6-react-a11y.yml` configuration. It requires the additional modules:
+
+- `eslint-plugin-react`
+- `eslint-plugin-jsx-a11y`
 
 ## Global Names and Environments
 
--- a/es6/mobify-es6-react-a11y.yml
+++ b/es6/mobify-es6-react-a11y.yml
@@ -1,0 +1,9 @@
+extends:
+  - './mobify-es6-react.yml'
+  - 'plugin:jsx-a11y/recommended'
+plugins:
+  - jsx-a11y
+rules:
+  # JSX A11y rules
+  jsx-a11y/href-no-hash: [1, 'Link']
+  jsx-a11y/img-has-alt: [1, 'Image']
--- a/es6/mobify-es6-react.yml
+++ b/es6/mobify-es6-react.yml
@@ -48,5 +48,5 @@ rules:
       argsIgnorePattern: '^(props|dispatch|state)$'
 
   # JSX A11y rules
-  jsx-a11y/href-no-hash: ['Link']
-  jsx-a11y/img-has-alt: ['Image']
+  jsx-a11y/href-no-hash: [1, 'Link']
+  jsx-a11y/img-has-alt: [1, 'Image']
--- a/es6/mobify-es6-react.yml
+++ b/es6/mobify-es6-react.yml
@@ -46,3 +46,6 @@ rules:
     - error
     - varsIgnorePattern: '^(styles)$'
       argsIgnorePattern: '^(props|dispatch|state)$'
+
+  # JSX A11y rules
+  jsx-a11y/no-static-element-interactions: off
--- a/es6/mobify-es6-react.yml
+++ b/es6/mobify-es6-react.yml
@@ -1,7 +1,6 @@
 extends:
   - './mobify-es6.yml'
   - 'plugin:react/recommended'
-  - 'plugin:jsx-a11y/recommended'
 globals:
   React: false
 parserOptions:
@@ -10,7 +9,6 @@ parserOptions:
     jsx: true
 plugins:
   - react
-  - jsx-a11y
 rules:
   # React error checking (beyond the default):
   react/no-render-return-value: error
@@ -46,7 +44,3 @@ rules:
     - error
     - varsIgnorePattern: '^(styles)$'
       argsIgnorePattern: '^(props|dispatch|state)$'
-
-  # JSX A11y rules
-  jsx-a11y/href-no-hash: [1, 'Link']
-  jsx-a11y/img-has-alt: [1, 'Image']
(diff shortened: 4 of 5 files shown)


marlowpayne:
Just ran this in Crabtree. Pretty cool stuff. It integrates very well with our existing eslint setup for Progressive builds and (for me) Atom text editor:
pr129atom

It also catches and displays linting errors as part of `npm run lint`:
pr129linter

A couple points:
- `eslint-plugin-jsx-a11y` needs to be installed locally as a devDep in the project (no biggie)
- A lot of our existing project and SDK components currently don't pass and will need fixing up (the SDK alone has 27 errors)

ericmuyser:
## Supported Rules

1. [anchor-has-content](docs/rules/anchor-has-content.md): Enforce all anchors to contain accessible content.
1. [aria-props](docs/rules/aria-props.md): Enforce all `aria-*` props are valid.
1. [aria-proptypes](docs/rules/aria-proptypes.md): Enforce ARIA state and property values are valid.
1. [aria-role](docs/rules/aria-role.md): Enforce that elements with ARIA roles must use a valid, non-abstract ARIA role.
1. [aria-unsupported-elements](docs/rules/aria-unsupported-elements.md): Enforce that elements that do not support ARIA roles, states, and properties do not have those attributes.
1. [click-events-have-key-events](docs/rules/click-events-have-key-events.md): Enforce a clickable non-interactive element has at least one keyboard event listener.
1. [heading-has-content](docs/rules/heading-has-content.md): Enforce heading (`h1`, `h2`, etc) elements contain accessible content.
1. [href-no-hash](docs/rules/href-no-hash.md): Enforce an anchor element's `href` prop value is not just `#`.
1. [html-has-lang](docs/rules/html-has-lang.md): Enforce `<html>` element has `lang` prop.
1. [img-has-alt](docs/rules/img-has-alt.md): Enforce that `` JSX elements use the `alt` prop.
1. [img-redundant-alt](docs/rules/img-redundant-alt.md): Enforce `` alt prop does not contain the word "image", "picture", or "photo".
1. [label-has-for](docs/rules/label-has-for.md): Enforce that `<label>` elements have the `htmlFor` prop.
1. [lang](docs/rules/lang.md): Enforce lang attribute has a valid value.
1. [mouse-events-have-key-events](docs/rules/mouse-events-have-key-events.md): Enforce that `onMouseOver`/`onMouseOut` are accompanied by `onFocus`/`onBlur` for keyboard-only users.
1. [no-access-key](docs/rules/no-access-key.md): Enforce that the `accessKey` prop is not used on any element to avoid complications with keyboard commands used by a screenreader.
1. [no-marquee](docs/rules/no-marquee.md): Enforce `<marquee>` elements are not used.
1. [no-onchange](docs/rules/no-onchange.md): Enforce usage of `onBlur` over `onChange` on select menus for accessibility.
1. [no-static-element-interactions](docs/rules/no-static-element-interactions.md): Enforce non-interactive elements have no interactive handlers.
1. [onclick-has-focus](docs/rules/onclick-has-focus.md): Enforce that elements with `onClick` handlers must be focusable.
1. [onclick-has-role](docs/rules/onclick-has-role.md): Enforce that non-interactive, visible elements (such as `<div>`) that have click handlers use the role attribute.
1. [role-has-required-aria-props](docs/rules/role-has-required-aria-props.md): Enforce that elements with ARIA roles must have all required attributes for that role.
1. [role-supports-aria-props](docs/rules/role-supports-aria-props.md): Enforce that elements with explicit or implicit roles defined contain only `aria-*` properties supported by that `role`.
1. [scope](docs/rules/scope.md): Enforce `scope` prop is only used on `<th>` elements.
1. [tabindex-no-positive](docs/rules/tabindex-no-positive.md): Enforce `tabIndex` value is not greater than zero.

marlowpayne:
Rule 3 seems to have issues with props set conditionally via JS.

```
aria-selected={`${activeIndex === idx ? 'true' : 'false'}`}
```

is considered a lint.

ericmuyser:
What about this?
```
aria-selected={activeIndex === idx ? 'true' : 'false'}
```

ericmuyser:
Just talked with Marlow. Based on the web dev sync, the purpose of this ticket is also to fool around with the rules to see if it won't cause problems for us later. There's some mystery around how far it goes down into the AST, like for conditionals in strings.

marlowpayne:
I set up this PR in the SDK to test/fix linting related to this code style addition: https://github.com/mobify/progressive-web-sdk/pull/168

marlowpayne:
Re. the [no-static-element-interactions](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md) rule: I feel like we quite commonly place interactive event handlers (`onClick`) on elements like `<div>`s, so I'm leaning towards removing this warning for our use.

jeffkamo:
@marlowpayne In those cases, should we actually be evaluating why we're using non-interactive elements in those situations? For example I believe we put a click handler on the accordion head element, which is a `div`. Instead of disabling the rule, should we actually make the accordion head element a `button`?

marlowpayne:
@jeffkamo that's a good point. That rule violation also comes up as a warning instead of a strict error. Which could be a good indicator that one should re-think the elements involved.

My only current concerns in our SDK components are a couple cases where it seems awkward to change the element due to semantics, but let's leave that discussion for that PR.

Right now I'll just turn back on that rule so we'll get the warnings in the SDK PR.

jeffkamo:
Note to self:

- [ ] I think `href-no-hash` needs to be customized to look at the `Link` component, since the docs don't suggest it looks at it by default like some other rules
- [ ] We should add a custom rule to `img-has-alt` to check our `Image` component too. As a side note, I'm surprised `img-has-alt` allows blank alt text (``) to pass... but then our propType rules can hopefully catch those.

jeffkamo:
Otherwise all the rules look desirable to have enabled to me.

jeffkamo:
Does anyone have any further thoughts on this ruleset? Is there anything blocking us from merging this?

marlowpayne:
I had a general thought re. these changes.

These changes are potentially lots of extra work for Progressive Web builds. When they install the next release of code style that contains this linter addition, if they aren't on a clean project they face potentially hours of refactoring or littering their code with `eslint-disable` statements. 

A couple options:
- Make this at least a Minor version bump, potentially Major
- Add a new `es6/mobify-es6-react-with-a11y.yml` config that simply extends `es6/mobify-es6-react.yml` with the additional a11y linting features. This means that existing builds can continue on the latest version of code-style (and any future enhancements to the config) without extensive efforts to address the a11y linter. New projects can be kicked off with this a11y-enhanced config from the start.

Maybe I'm overthinking this, especially considering how much we would like to have a11y in the forefront (maybe we _should_ force a11y lint fixes...). Would love some other thoughts.

jeffkamo:
I hear what you're saying, and that's a pretty fair concern. In regards to creating an extendable config file, does it need its own file? Can't a project extend this same file and apply/override rules as necessary there in the same way as normal?

marlowpayne:
@jeffkamo yes, what I'm proposing is a separate file e.g. `es6/mobify-es6-react-with-a11y.yml`. It would be something like this:
```
extends:
  - './node_modules/mobify-code-style/es6/mobify-es6-react.yml'
  - 'plugin:jsx-a11y/recommended'
plugins:
  - jsx-a11y
rules:
  # JSX A11y rules
  jsx-a11y/href-no-hash: [1, 'Link']
  jsx-a11y/img-has-alt: [1, 'Image']
```

Projects then would extend this with their `.eslintrc.yml`:
```
extends:
  - './node_modules/mobify-code-style/es6/mobify-es6-react-with-a11y.yml'
```

> Can't a project extend this same file and apply/override rules as necessary there in the same way as normal?

Yes, while this is possible, in its current form, it's difficult to use the "extends" pattern correctly for the case I described. Because the project extending `es6/mobify-es6-react.yml` and then _removing_ an eslint plugin is not an "extension". Extensions should be additive, with each extension adding its own distinct functionality. For example: `es6/mobify-es6-react.yml` extends `es6/mobify-es6.yml` which extends eslint's `eslint:recommended` ruleset. Each one builds on the previous.

jeffkamo:
@marlowpayne Ah okay, gotcha. I've added a todo to move these rules into its own file.

marlowpayne:
I believe this is good to go, but let's get _one_ more set of 👀  on this before we merge and publish.

MikeKlemarewski:
Tested it out and it works well.
👍  if we're happy with these rules.

PR #130, opened 2016-09-05
Add Markdown Linting

johnboxall:
Standardizes Mobify's markdown formatting using [`remark-lint`](https://github.com/wooorm/remark-lint).

For the most part, I think this is ready to go. Outstanding:
- Is this the formatting we want?
- Do we want to specify the dependancies any other way?
- Is there a more robust way to find `remark-cli`? The current implementation will break with `npm3`'s shallow dependancies.
- ...?
## Changes
- Adds a utility script for easily incorporating markdown linting into repos.
## How To Test
- `npm test`
## TODOs:
- [ ] +1
- [ ] Updated README
- [ ] Updated CHANGELOG


docs/README.md
```diff
@@ -1,3 +1,32 @@
...
+
+To make writing document easier, we provide:
+
+* A linter for consistent markdown style.
+* A writing checklist for consistent process.
+
+### Markdown Linting
+
+```bash
+npm install --save mobify-code-style
```
jeremywiebe: `--save-dev` ? This isn't a runtime dependency, is it?

docs/README.md
```diff
@@ -1,3 +1,32 @@
+> Write great documentation! – * [_Mobify Developer Values_](https://github.com/mobify/developer-values#️-write-great-documentation)
+
+At Mobify, documentation is written in markdown.
+
+Documentation should live as close as possible to the code it documents.
+Typically, this means including it in a `docs` folder at the root of the repo.
```
jeremywiebe: The script in the docs below assumes a `/docs` folder. Is there any repo that this isn't true for right now?  Maybe we should just say that docs _should_ be in a `/docs` folder. #convention-over-configuration :)

package.json
```diff
@@ -2,10 +2,6 @@
   "name": "mobify-code-style",
   "version": "2.4.3",
   "description": "Code style guide and linting tools for Mobify",
-  "main": "index.js",
-  "scripts": {
-    "test": "grunt test"
```
jeremywiebe: Weird. I can't even find a `Gruntfile` in this repo, so this doesn't appear to have done anything.

benlast:
I think a neat way to deliver this would be to create a Dockerfile based on one of the `node` images (https://hub.docker.com/_/node/), containing the commands to install the linter, with the CMD being a bash (or node) script that takes one or more filenames to check. Then a `docker run` command could lint markdown files in the current directory.

jeremywiebe:
Pairing with @johnboxall   Verbal +1

PR #134, opened 2016-09-28
Update styleguide for swift

kerrmarin:
Fixed a couple examples that wouldn't compile, and chucked in what I think we should do with optionals ❓ ❔  😆 

Edit: I've updated this with a reference from https://github.com/raywenderlich/swift-style-guide

### DIFF
--- a/swift/README.md
+++ b/swift/README.md
@@ -58,7 +58,6 @@ Our overarching goals are conciseness, readability, and simplicity.
 
 Consider warnings to be errors. This rule informs many stylistic decisions such as not to use deprecated methods, C-style for loops, or strings as selectors.
 
-
 ## Naming
 
 Clarity at the point of use is your most important goal. Entities such as methods and properties are declared only once but used repeatedly. Design APIs to make those uses clear and concise. When evaluating a design, reading a declaration is seldom sufficient; always examine a use case to make sure it looks clear in context.
--- a/swift/README.md
+++ b/swift/README.md
@@ -9,7 +9,8 @@ Prefer interpolation over concatenation:
 
 ```swift
 let name = "John"
-let messsage = "Hello \(name)"
+let messsage = "Hello \(name)" //YES
+let greeting = "Hello " + name //NO
 ```
 
 Optionals
@@ -19,21 +20,46 @@ Use `as` for type coercion if possible (this is enforced statically). Otherwise 
 
 DO NOT use `as!` or `value!` because Xcode told you so. Stop. Think. You probably want to use `if let value = value`.
 
-You should use `!` only if you just assigned to an object, you know initialization can not fail, or you are initializing an object during init() but need to pass `self` to another object's init():
+You should use `!` only if:
+
+- You just assigned to an object where you know initialization can not fail:
 
 ```swift
-let regex = NSRegularExpression(pattern: "[a-z]", options: NSRegularExpressionOptions.allZeros, error: nil)!
+let numberSeven = Int("7")! //Don't do this.
 ```
 
-init() example:
+In this case, wrap this into a static function and write a test for it:
+
 ```swift
-class Component {
+extension Int {
+    static func seven() -> Int {
+        return Int("7")!
+    }
+}
+```
+
+```swift
+import XCTest
+
+class IntegerTests: XCTestCase {
+    func testSeven() {
+        XCTAssertNotNil(Int.seven())
+    }
+}
+```
+
+Although the test might seem redundant, because `seven()` returns a non-optional, this ensures that even if in the future someone changes the implementation of `seven()`, it won't get past CI if the explicit unwrapping would fail.
+
+- You are creating an object during initialization but need to pass `self` to that object's initializer:
+
+```swift
+class Component: ParentComponent {
     var controller: CustomViewController! // We want to use it as a non-optional but have to initialize after super.init()
-    
-    init() {
+
+    override init() {
         // controller = CustomViewController(component: self) <-- compiler error
         super.init()
-        controller = CustomViewController(component: self) 
+        controller = CustomViewController(component: self)
     }
 }
 ```
@@ -42,11 +68,6 @@ How do I handle errors in Swift?
 ================================
 
 As of Swift 2.0 exceptions have been introduced into the language and the APIs that formerly used `someMethod(..., error: *NSError)` make use of them. We have decided to stick with our existing pattern for returning errors.
-
-You have a couple options:
-
-- Return a tuple (similar to Go)
-- Use a result/error enumeration type
 
 We typically use an enumeration type:
 
--- a/swift/README.md
+++ b/swift/README.md
@@ -93,7 +93,7 @@ class Component {
 }
 ```
 
-Use calculated properies instead of getter and setter functions:
+Use computed properties instead of getter and setter functions:
 
 ```swift
 class Component {
@@ -144,7 +144,7 @@ Try to wrap comments to 100 characters, though.
 For Dictionary and Array literals, use python-style indentation:
 
 ```swift
-let nameMap: [String: String] = [
+let nameMap = [
     "George": "Jetson",
     "Astro": "Boy",
     // etc.
@@ -217,7 +217,7 @@ Avoid this design pattern.
 
 Use `let` unless the variable will be mutated.
 
-delegates
+Delegates
 =========
 
 Use the `weak` modifier for `delegate` properties.  
@@ -354,12 +354,12 @@ Group helper functions into a `struct` rather than a `class`.
 Extensions
 ==========
 
-Create `extension`s to classes when you will use those functions in alot of places.  Otherwise use [Helper functions](#Helper functions).  
+Create `extension`s to classes when you will use those functions in a lot of places. Otherwise use [Helper functions](#Helper functions).  
 
 Layout
 ======
 
-Prefer Autolayout over springs+struts (autoresizing mask).  Autolayout automatically handles many things that springs+struts doesn't (status bar hiding/showing, device rotation)
+Prefer Autolayout over springs+struts (autoresizing mask). Autolayout automatically handles many things that springs+struts doesn't (status bar hiding/showing, device rotation)
 
 Other useful patterns
 =====================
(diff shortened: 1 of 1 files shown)


swift/README.md
```diff
@@ -369,25 +369,25 @@ Generic functions:
...
+        return nil
     }
 
-    return nil
+    return rightType
 }
 ```
 
-Note that the type T is inferred by what you assign `getValue()` to.  So in this case getValue() infers String? because `message` is a String?.
+Note that the type `T` is inferred by what you assign `getValue()` to.  So in this case getValue() infers `String?` because `message` is a `String?`.
```
coffee-cup: 👾 There are two spaces before the word 'So' :)

kerrmarin: Fixed that now 💯

swift/README.md
```diff
@@ -19,21 +20,46 @@ Use `as` for type coercion if possible (this is enforced statically). Otherwise
...
+    }
+}
+```
+
+```swift
+import XCTest
+
+class IntegerTests: XCTestCase {
+    func testSeven() {
+        XCTAssertNotNil(Int.seven())
```
jeremywiebe:
```swift
XCTAssertEqual(7, Int.seven())
```
?

swift/README.md
jeremywiebe: @coffee-cup Back in the day that's how keyboarding was taught (two spaces after a period).  You can still find holy wars about that on the internet. :)

swift/README.md
kerrmarin: 👍

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+
+Sometimes, things should be structs but need to conform to `AnyObject` or are historically modeled as classes already (`NSDate`, `NSSet`). Try to follow these guidelines as closely as possible.
+
+### Example definition
+
+Here's an example of a well-styled class definition:
+
+```swift
+class Circle: Shape {
+  var x: Int, y: Int
```
jeremywiebe: Might be just me, but I'm not a fan of multiple properties defined on the same line.

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+
+Classes have [reference semantics](https://developer.apple.com/library/mac/documentation/Swift/Conceptual/Swift_Programming_Language/ClassesAndStructures.html#//apple_ref/doc/uid/TP40014097-CH13-XID_145). Use classes for things that do have an identity or a specific life cycle. You would model a person as a class because two person objects are two different things. Just because two people have the same name and birthdate, doesn't mean they are the same person. But the person's birthdate would be a struct because a date of 3 March 1950 is the same as any other date object for 3 March 1950. The date itself doesn't have an identity.
+
+Sometimes, things should be structs but need to conform to `AnyObject` or are historically modeled as classes already (`NSDate`, `NSSet`). Try to follow these guidelines as closely as possible.
+
+### Example definition
+
+Here's an example of a well-styled class definition:
```
jeremywiebe: So, I'm a bit conflicted if we need this. At the top we state that the Astro codebase is the canonical example of Swift code. I wonder if this is then redundant.

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+class TestDatabase : Database {
+  var data :[String:CGFloat] = ["A" : 1.2, "B":3.2]
+}
+```
 
 For Dictionary and Array literals, use python-style indentation:
 
 ```swift
-let nameMap: [String: String] = [
+let nameMap = [
```
jeremywiebe: Good catch!   I wonder if when we wrote this the Swift compiler wasn't smart enough to infer this.

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
 
-Line lengths
-============
+* Colons always have no space on the left and one space on the right. Exceptions are the ternary operator `? :`, empty dictionary `[:]` and  `#selector` syntax for unnamed parameters `(_:)`.
 
-You are going to have long lines in Swift. Deal with it.
+**Preferred:**
+```swift
+class TestDatabase: Database {
+  var data: [String: CGFloat] = ["A": 1.2, "B": 3.2]
```
jeremywiebe: Or this....
```swift
var data = ["A": 1.2, "B": 3.2]
```

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
-Instead of:
+Since the compiler does not allow you to re-declare protocol conformance in a derived class, it is not always required to replicate the extension groups of the base class. This is especially true if the derived class is a terminal class and a small number of methods are being overriden. When to preserve the extension groups is left to the discretion of the author.
+
+For UIKit view controllers, consider grouping lifecycle, custom accessors, and IBAction in separate class extensions.
+
+### Unused Code
 
+Unused (dead) code, including Xcode template code and placeholder comments should be removed. Aspirational methods whose implementation simply calls the super class should also be removed.
+
+**Not Preferred:**
```
jeremywiebe: We revert the order of Not Preferred and Preferred with this example. Is that intentional?

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+>
+> If you call `dateFromString(_:)` make sure that you provide a string with the format "yyyy-MM-dd".
+>
+> If you call `timedAction(afterDelay:perform:)` from `viewDidLoad()` remember to provide an adjusted delay value and an action to perform.
+>
+> You shouldn't call the data source method `tableView(_:cellForRowAtIndexPath:)` directly.
+
+This is the same as the `#selector` syntax. When in doubt, look at how Xcode lists the method in the jump bar – our style here matches that.
+
+### Class Prefixes
```
jeremywiebe: I'm tempted to say we don't need this. I can't think of one time where we ran into this at Mobify. It is probably more of an issue for engineers coming from an Obj-C background where class prefixing was more of a "thing".

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+* [String Interpolation](#string-interpolation)
+* [Objective-C Interoperability](#objective-c-interoperability)
+* [Typealiasing](#typealiasing)
+* [Tests](#tests)
+* [Autolayout](#autolayout)
+* [Credits](#credits)
+
+## Correctness
+
+Consider warnings to be errors. This rule informs many stylistic decisions such as not to use deprecated methods, C-style for loops, or strings as selectors.
```
kerrmarin: It is :)

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+  manager.delegate = self
+  manager.requestAlwaysAuthorization()
+  return manager
+}()
+```
 
-Comments
-========
+**Notes:**
+  - `[unowned self]` is not required here. A retain cycle is not created.
```
kerrmarin: Sure. Technically a retain cycle is created (but its super short-lived, so I don't count it). The reason it doesn't create a permanent retain cycle is because (as you say) the closure is executed once _and_ because the value assigned to the lazy var is the manager. A simple way to make a retain cycle here would be to change this to be:

```swift
lazy var locationManager: () -> CLLocationManager = { // [unowned self] in <- would break the retain cycle here
   let manager = CLLocationManager()
   manager.desiredAccuracy = kCLLocationAccuracyBest
   manager.delegate = self
   manager.requestAlwaysAuthorization()
   return manager
 }
```
this would create a permanent retain cycle because it's the closure that's being set to the lazy var, not the value that the closure returns, and the closure captures `self`. I'll add an explanation to the docs.

swift/README.md
kerrmarin: 👍  I agree. I've changed this to be in two different lines.

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
 
-No single line `if...else` statements
+## Types
+
+Always use Swift's native types when available. Swift offers bridging to Objective-C so you can still use the full set of methods as needed.
+
+**Preferred:**
+```swift
+let width = 120.0                                    // Double
+let widthString = (width as NSNumber).stringValue    // String
```
kerrmarin: I've changed the names of the variable -- can you think of a better example of bridging? I can't think of anything better than this 🤔

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
-Name the delegate property `delegate` unless your class supports multiple different delegates.  
+When naming optional variables and properties, avoid naming them like `optionalString` or `maybeView` since their optional-ness is already in the type declaration.
 
-Protocols (aka Interfaces)
-==========================
+For optional binding, shadow the original name when appropriate rather than using names like `unwrappedView` or `actualLabel`.
 
-Protocols are awesome. Use them to restrict the API surface area of an object
-being passed around.
+Use `as` for type coercion if possible (this is enforced statically). Otherwise use `as?`.
```
kerrmarin: Umm, I think the difference is between guaranteed conversion (`as`, upcasting, verified by the compiler that it will succeed) and forced conversion (`as?`, `as!`, downcasting, can't be verified by the compiler). So this rule is intended to be "use `as` when possible (upcasting), if you can't use `as`, use `as?` (for downcasting), but don't use `as!`. Maybe I can make it a little clearer?

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+let currentBounds: CGRect = computeViewBounds()
+let names = [String]()
 ```
 
-`if`/`else`
-===========
+#### Type Annotation for Empty Arrays and Dictionaries
 
-This is the style:
+For empty arrays and dictionaries, use type annotation. (For an array or dictionary assigned to a large, multi-line literal, use type annotation.)
```
kerrmarin: Good idea 👍

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
-Extensions
-==========
+Swift does not require a semicolon after each statement in your code. They are only required if you wish to combine multiple statements on a single line.
 
-Create `extension`s to classes when you will use those functions in alot of places.  Otherwise use [Helper functions](#Helper functions).  
+Do not write multiple statements on a single line separated with semicolons.
 
-Layout
-======
+The only exception to this rule is the `for-conditional-increment` construct, which requires semicolons. However, alternative `for-in` constructs should be used where possible.
```
kerrmarin: Good catch!

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+  self?.updateUI(model)
 }
 ```
 
-`if let` variables
-==================
+## Access Control
 
-Use shadowing:
+Full access control annotation in tutorials can distract from the main topic and is not required. Using `private` appropriately, however, adds clarity and promotes encapsulation. Use `private` as the leading property specifier. The only things that should come before access control are the `static` specifier or attributes such as `@IBAction` and `@IBOutlet`.
```
kerrmarin: I've pushed changes to that :)

jeremywiebe:
👍    Nice work!

swift/README.md
```diff
@@ -1,372 +1,1100 @@
...
+
+## Memory Management
+
+Code (even non-production) should not create reference cycles. Analyze your object graph and prevent strong cycles with `weak` and `unowned` references. Alternatively, use value types (`struct`, `enum`) to prevent cycles altogether.
+
+Variables that are delegates should be `weak`.
+
+### Extending object lifetime
+
+Extend object lifetime using the `[weak self]` and `guard let strongSelf = self else { return }` idiom. `[weak self]` is preferred to `[unowned self]` where it is not immediately obvious that `self` outlives the closure. Explicitly extending lifetime is preferred to optional unwrapping.
```
jeremywiebe: Yes, that makes sense.  I think when I read this I was focusing on the `[weak self]` bit.  Wonder if we could re-word this.  I think it's basically two things we need to outline:
1. Use `[weak self]` to avoid a retain cycle.
2. Within a block, create a reference to `self` immediately and use that reference throughout the block. This avoids `self` (which is weak) being released during execution of the block.

kerrmarin: I've changed this now: https://github.com/mobify/mobify-code-style/pull/134/commits/20237332b7c48ac30af85d823fa87ed4f522153b

jeremywiebe:
👍  (for realz now!)  🎉

PR #139, opened 2016-12-01
Add note about PyLint

benlast:
## Changes
- Add note to README for Python to explain that PyLint is slow, and Flake8 is an acceptable level of linting.


sketchytechky:
👍 Let's ship it

PR #150, opened 2017-02-02
Python docstring format recommendation

benlast:
While working on [one of my own OSS projects](https://github.com/benlast/grimace), I researched docstring comment standards and how widely used they are. I though Google's was the simplest, and quite widely used, so I'm recommending it. The major alternative would be ReStrucuredText, but the learning curve for that is significantly higher, and I think there are reasonable arguments for avoiding it (see [this](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/)).

## Changes
- Add a recommendation for a Python docstring commenting standard, with links and example.

## TODOs:
- [ ] +1

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -148,3 +148,14 @@ Message codes can be found [here](http://pylint-messages.wikidot.com/all-codes).
 Disable works for the block in which they are found, so include it at the module
 level to disable a message for a module or file.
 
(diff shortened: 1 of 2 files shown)


olibrook:
👍  From me, I'd like to see something in there about Python 3 type annotations in future though.

PR #151, opened 2017-02-02
Fix doc linting in node 6

jeremywiebe:
Linked PRs: N/A

## Changes
- In node 4 and earlier `node_modules` are nested (each module has its own `node_modules` folder with its dependencies.  In node 6 all dependencies are in a flattened, top-level `node_modules`.  This PR fixes the `lint-md` command so that it can always find the `remark` binary.

## How To Test
- Reference this branch of `mobify-code-style` in the Astro project
- Switch to node 4 and run `rm -rf node_modules; npm install && npm run docs:lint`
- Switch to node 6 and run `rm -rf node_modules; npm install && npm run docs:lint`
- Both will result in a ton of linting errors, but you should not see a node error saying it cannot find `remark`

- N/A

## TODOs:
- [ ] +1
~~- [ ] Updated README~~
~~- [ ] Updated CHANGELOG~~
~~- [ ] (Other applicable TODOs)~~

### DIFF
--- a/docs/bin/lint-md
+++ b/docs/bin/lint-md
@@ -14,11 +14,16 @@ done
 DOCS_DIR="$(cd "$(dirname "$BIN")/.." && pwd)"
 REMARK_RC="$DOCS_DIR/remarkrc"
 
-# This is tricky. `remark` is a dep of mobify-code-style
-# and so won't show up in `node_modules/.bin` of a project
-# that uses mobify-code-style. We have to move into the
-# the mobify-code-style /docs directory and find the npm
-# bin directory relative that _that_ to find remark!
-REMARK_CLI="$(cd "$DOCS_DIR" && npm bin)/remark"
+if [[ $(node -v) =~ ^v4 ]]; then
+	# Node v4!!
+	# This is tricky. `remark` is a dep of mobify-code-style
+	# and so won't show up in `node_modules/.bin` of a project
+	# that uses mobify-code-style. We have to move into the
+	# the mobify-code-style /docs directory and find the npm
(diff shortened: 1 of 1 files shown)


docs/bin/lint-md
```diff
@@ -14,11 +14,16 @@ done
...
-# bin directory relative that _that_ to find remark!
-REMARK_CLI="$(cd "$DOCS_DIR" && npm bin)/remark"
+if [[ $(node -v) =~ ^v[0-4] ]]; then
+	# Node v4!!
+	# This is tricky. `remark` is a dep of mobify-code-style
+	# and so won't show up in `node_modules/.bin` of a project
+	# that uses mobify-code-style. We have to move into the
+	# the mobify-code-style /docs directory and find the npm
+	# bin directory relative that _that_ to find remark!
+    REMARK_CLI="$(cd "$DOCS_DIR" && npm bin)/remark"
```
benlast: Looks reasonable. An alternative would be to use `find` to walk down the directory tree, locate all `bin` directories and invoke the first `remark` that it finds in one :) But that's just an observation...

benlast:
@jeremywiebe assuming you tested this, it looks like a reasonable solution :) 👍

benlast:
oops

docs/bin/lint-md
jeremywiebe: @benlast I'm hoping the v4 branch of this script begins to rot very fast and we can just delete it in the not-to-distant future.  This part of the branch is already working so I'd prefer to leave it if that's ok with you.

benlast: Fine by me - it was only an observation anyway :)

PR #153, opened 2017-02-22
Adding a small section on ES6 naming

jeremywiebe:
Linked PRs: N/A

## Changes
- Added a section to the ES6 code style to clarify casing in names

## How To Test
- Review updated sections
- Review `new-cap` rule. I'm thinking we can simplify the YAML config, but this is what I could get to work. 

- [`new-cap` rule](http://eslint.org/docs/rules/new-cap.html)
- [AirBNB Code Style 22.2, 22.3](https://github.com/airbnb/javascript#naming--camelCase)

## TODOs:
- [ ] +1
- [ ] Updated README
- [ ] Updated CHANGELOG


es6/README.md
```diff
@@ -51,6 +51,35 @@ globals:
 The boolean value indicates whether the global is to be treated as
 read-only (`false`) or read-write (`true`).
 
+## Naming
+
+Classes and constructor functions should be TitleCased
```
jeremywiebe: So one thing that violates this now is [`mobify-debug-js`](https://github.com/mobify/mobify-debug-js/blob/master/index.js#L19).

I _think_ that should become a lowercase function name... thoughts?

es6/README.md
```diff
@@ -51,6 +51,35 @@ globals:
...
+// GOOD
+const defaultName = 'Element'
+
+class MyClass {
+    const name = 'Button'
+    render() { ... }
+}
+
+// BAD - ES6 has `const` now. The compiler/interpreter will warn you if you re-assign a const.
+const DEFAULT_NAME = 'Element'
```
DanielBrookRoberge: The `const` keyword is very useful for indicating things, but it doesn't replace `ALL_CAPS_NAMING` entirely. We get errors for re-assigning a const primitive, but it's different for a composite/mutable value such as an array or object.

Basically, if I see, at the module level,
```js
const stateCallbacks = {
    pending: sendPendingResponse,
    completed: sendCompletedResponse,
    archived: sendArchivedResponse
}
```
then I'd expect that reassigning the members of it would be an intended thing, while if we have 
```js
const STATE_CALLBACKS = {
    pending: sendPendingResponse,
    completed: sendCompletedResponse,
    archived: sendArchivedResponse
}
```
then I'll expect that the contents are intended not to change.

es6/README.md
```diff
@@ -51,6 +51,56 @@ globals:
...
+
+```javascript
+// GOOD
+class MyClass {
+    const name = 'Button'
+    render() { ... }
+}
+
+// BAD - `const` outside of a method should be ALL_CAPS
+const defaultName = 'Element'
```
jeremywiebe: As I'm re-reading this, we don't follow this in the `platform-scaffold`.  We export many const symbols that are not `SCREAMING_CAPS`, but they're usually functions. I wonder if the distinction of when to use `SCREAMING_CAPS` is that the value is "data" (ie. it's not a function):

```js
// GOOD - exporting a function
export const receiveProductData = createAction('Receive Product Data')

// BAD - exporting _data_
export const defaultName = 'Element'
```

jansepar: It would be good to have a GOOD and a BAD for "exporting a function", and GOOD/BAD for "const outside of a method"

jansepar:
👍

PR #155, opened 2017-05-03
Copyright Headers Tool

Cdok:
## Changes
- This PR adds a tool to either check for existing copyright headers, or add copyright headers if they are missing
- This tool is language agnostic (see note below), and can be pointed to any directory using a glob pattern

**Note:** this tool reads from the `headers/copyright.extension` files to select what header to insert, where `extension` is the target language

Currently only `.py`, `.jsx`, and `.js` have been added for testing.

## How To Test

- Clone the repo and checkout this branch (`copyright-headers`)
- `cd` into the `copyright` directory, and run `node copyright.js 'PATTERN'` where `PATTERN` is a glob pattern
- For example, if you cloned this repo in the same directory as the `progressive-web-sdk`, and needed to add copyright headers to all `.js` files in the `src` directory of the SDK, you would write:

```bash
node copyright.js '../../progressive-web-sdk/src/**/*.js' --fix
```

Note the optional `--fix` flag that can be passed in, which will add the headers to files which do not contain it.

Another flag `--update` will check the target files for outdated headers, and update them to the current year.

The default mode (no flags) will exit the process with an error if any of the targeted files are missing copyright headers.

### To test in dev mode (with chrome inspect)
`node --inspect --debug-brk copyright.js ../../progressive-web-sdk/bin/**/*.js`

## TODOs:
- [x] +1
- [x] Updated README
- [x] Updated CHANGELOG

## Resources:
[node-glob](https://github.com/isaacs/node-glob)


README.md
```diff
@@ -104,3 +104,10 @@ instructions on using `pep8`, `pylint` and `pyflakes` to check Python code.
 ## Documentation
 
 See the [`docs`](docs) directory for setup and configuration of Markdown linting.
+
+
+## Copyright Headers
+
+All Mobify source files should contain copyright headers at the beginning of the file.
```
jeremywiebe: @johnboxall "*All*" or just source we intend to ship outside our control (ie. to the browser/device). I wouldn't anticipate server code needs these?

copyright/README.md
```diff
@@ -0,0 +1,30 @@
+## Mobify Copyright Headers
+
+This tool reads from the `headers/copyright.extension.txt` text files to select what header to insert, where `extension` is the target file extension
+
+## Quick Start
+
+- Clone this repo
```
jeremywiebe: I've seen these formatted as a short terminal session... sometimes that's clearer.  Eg.

```sh
git clone git@github.com:mobify/mobify-code-style.git
cd mobify-code-style/copyright
node copyright.js ../**/*.js ../**/*.py
```

copyright/headers/copyright.js.txt
```diff
@@ -0,0 +1,3 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
```
jeremywiebe: Can we make the year dynamic?  It'd be a pain to have to update this every year. :)  When we run the tool it asserts the copyright as of the current year, I think that'd be enough.

package.json
```diff
@@ -19,6 +19,7 @@
     "test": "cd docs/test; bash test.sh"
   },
   "dependencies": {
+    "glob": "^7.1.1",
```
jeremywiebe: Can we add this tool to the `"bin"` key just a few lines above?  That will make it much easier to use in projects that import this npm package.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+    missingHeaders: [],
+    copyrightText: '',
+    lintMode: false,
+    updateHeaders() {
+        const filesRead = this.items.map((item) => {
+            return new Promise((resolve) => {
+                fs.readFile(item, (err, data) => {
+                    if (!this.hasCopyrightHeader(data)) {
+                        if (this.lintMode === true) {
+                            console.log(`\x1b[33m Missing copyright headers in \x1b[36m ${item}`)
```
jeremywiebe: It'd be good to use constants for these escape codes.  Without running the tool I have no idea what `\x1b[33m` does.  :)

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
```
jeremywiebe: :)  

imgres

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+            }
+        })
+
+    },
+    hasCopyrightHeader(data) {
+        if (data.indexOf('Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved.') >= 0) {
+            return true
+        }
+        return false
+    },
```
DanielBrookRoberge: This function can just be `return data.indexOf(COPYRIGHT_STRING) >= 0`.

It doesn't actually use `this`, so it could even move outside this object and become even simpler.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+const glob = require('glob')
+const fs = require('fs')
+
+const copyright = {
+    items: [],
+    missingHeaders: [],
+    copyrightText: '',
```
jeremywiebe: This assumes that the glob(s) you pass will only ever be of the same type. It might be better to keep a dict of extension -> header text so that you can evaluate on a per-file basis which header you need. 

Given that we'll probably only have a handful of file types supported, you could read those in before you start and have them handy throughout the run.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+        // Sets lint flag if the user provides --lint command line arg
+        if (args.indexOf('--lint') >= 0) {
+            args.splice(args.indexOf('--lint'), 1)
+            this.lintMode = true
+        }
+
+
+        const processedGlobs = args.map((dir) => {
+            return new Promise((resolve) => {
+                this.getHeaderText(dir)
```
DanielBrookRoberge: If the `dir`s have different header text, this will cause them to step on each other.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+                glob(`${dir}`, (err, files) => {
+                    this.items.push(files[0])
+                    resolve()
```
DanielBrookRoberge: Rather than using the `this.items` array, we can just pass `files[0]` to `resolve`. Then the `Promise.all` call below will resolve to an array of the same contents.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+        const ext = globStr.match(/\.[0-9a-z]+$/i)[0]
+        const path = `headers/copyright${ext}.txt`
+
+        if (!fs.existsSync(path)) {
+            console.log(`\x1b[31m \x1b[40mERROR\x1b[49m - ${path} does not exist`)
+            process.exit(1)
+        } else {
+            this.copyrightText = fs.readFileSync(path)
+        }
+    },
```
DanielBrookRoberge: With the files that are included here, we'll get an error for `.jsx` files, which are very common in the Progressive code.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+const glob = require('glob')
+const fs = require('fs')
+
+const copyright = {
+    items: [],
```
jeremywiebe: What are these items? :)  I think they're individual file paths, so renaming them to such would make this variable alot easier to "digest".  Maybe `filePaths: []` or even `filesToProcess: []`?

copyright/README.md
Cdok: @jeremywiebe so ideally would we want something like 
```bash
npm install mobify-code-style
npm run lint:copyright
```

Or something along these lines?

I made the change in the README right now - so I should also add something like

```bash
"lint:copyright": "node_modules/mobify-code-style/copyright/copyright.js PATH --lint"
```

to your `package.json` - or is there an easier way than digging into the node modules in the npm script 😛 🤔

Cdok:
@DanielBrookRoberge @jeremywiebe thank you 😄  addressed the issues with namespace pollution, and refactored into more of a pipeline.

Ready for more feedback 🎉

copyright/copyright.js
```diff
@@ -0,0 +1,109 @@
...
+            return fs.readFileSync(textPath).toString().replace('year', new Date().getFullYear())
+        }
+    },
+    buildSupportedExtensions() {
+        const supportedHeaders = path.join(__dirname, './headers')
+        fs.readdir(supportedHeaders, (err, filenames) => {
+            filenames.forEach((file) => {
+                const extension = file.match(/\.[0-9a-z]+/i)[0]
+                this.langs[extension] = file
+            })
```
DanielBrookRoberge: The header files don't really need to have a `.txt` extension; if we leave it off we can use the same extension extraction logic for both this and the actual source files.

copyright/copyright.js
```diff
@@ -0,0 +1,109 @@
...
+
+const copyright = {
+    lintMode: false,
+    langs: {},
+    run() {
+        this.buildSupportedExtensions()
+        const processedGlobs = args.map((folder) => {
+            return new Promise((resolve) => {
+                glob(`${folder}`, (err, file) => {
+                    resolve(file[0])
```
DanielBrookRoberge: This looks like it only gets the first filename matching the glob, is that what you want?

copyright/copyright.js
```diff
@@ -0,0 +1,109 @@
...
+                })
+            })
+        })
+
+        Promise.all(processedGlobs).then((files) => {
+            const filesMissingHeader = files.filter((file) => {
```
DanielBrookRoberge: This variable appears to contain filenames that do contain the header, rather than ones that are missing the header, right?

copyright/copyright.js
```diff
@@ -0,0 +1,109 @@
...
+                        return false
+                    } else {
+                        const newData = this.getHeaderText(ext) + content
+                        fs.writeFileSync(file, newData)
+                        console.log(`${green}Copyright header succesfully written into ${magenta}${file}`)
+                        return true
+                    }
+                } else {
+                    return true
+                }
```
DanielBrookRoberge: This could be simpler and less nested if you start out with 
```
if (hasCopyrightHeader) {
    return true
}
```

copyright/copyright.js
```diff
@@ -0,0 +1,109 @@
...
+                    resolve(file[0])
+                })
+            })
+        })
+
+        Promise.all(processedGlobs).then((files) => {
+            const filesMissingHeader = files.filter((file) => {
+                const content = fs.readFileSync(file)
+                const copyrightStr = 'Copyright (c)'
+                const hasCopyrightHeader = content.indexOf(copyrightStr) >= 0
```
DanielBrookRoberge: In Node, we can use `String.prototype.includes` for something like this.

copyright/copyright.js
Cdok: Each matched `file` that's inside `glob()` callback is an Array of length 1
screen shot 2017-05-04 at 10 51 27 am

From `node-glob`:

```javascript
var glob = require("glob")

// options is optional
glob("**/*.js", options, function (er, files) {
  // files is an array of filenames.
  // If the `nonull` option is set, and nothing
  // was found, then files is ["**/*.js"]
  // er is an error object or null.
})
```

So it seems like this should be an array of all of the matches - but instead the functionality is that the callback is fired 1 time for each matched file - and `file` is an array of length 1 🤔

DanielBrookRoberge: So then this wouldn't work with the `resolve` method since it would only grab the first result (subsequent calls to `resolve` do nothing). 

It might just be simpler to use `glob.globSync` and then flatten the resulting array of arrays.

copyright/copyright.js
```diff
@@ -0,0 +1,108 @@
...
+                    if (this.lintMode) {
+                        console.log(`${yellow}${file} ${red}missing copyright header`)
+                        return false
+                    } else {
+                        const newData = this.getHeaderText(ext) + content
+                        fs.writeFileSync(file, newData)
+                        console.log(`${green}Copyright header succesfully written into ${magenta}${file}`)
+                        return true
+                    }
+                }
```
DanielBrookRoberge: Having restructured things in this way, we don't actually need the `else` block; the `return` statement will exit the function.

package.json
```diff
@@ -12,7 +12,8 @@
     "url": "https://github.com/mobify/mobify-code-style/issues"
   },
   "bin": {
-    "lint-md": "docs/bin/lint-md"
+    "lint-md": "docs/bin/lint-md",
+    "glob": "^7.1.1"
```
jeremywiebe: Hrm... is this in the right place?  Isn't this a dependency?

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+const copyright = {
+    lintMode: true,
+    langs: {},
+    run() {
+        this.buildSupportedExtensions()
+        let files = []
+
+        const processedGlobs = args.map((folder) => {
+            files.push(glob.sync(folder))
+        })
```
DanielBrookRoberge: If you return `glob.sync(folder)` from the map function, then you won't need the `files` array; `processedGlobs` will contain what you want.

copyright/copyright.js
```diff
@@ -0,0 +1,106 @@
...
+                    console.log(`${yellow}${file} ${red}missing copyright header`)
+                    return false
+                } else {
+                    const newData = this.getHeaderText(ext) + content
+                    fs.writeFileSync(file, newData)
+                    console.log(`${green}Copyright header succesfully written into ${magenta}${file}`)
+                    return true
+                }
+            }
+        })
```
DanielBrookRoberge: Thinking about this some more, we don't really need this array, just a flag for whether an error has occurred. Then this can be a `forEach` that just sets the flag rather than worrying about return values.

Cdok:
Just added the node environment shebang, and updated the readme to reflect the new installation instructions.

On the thought of shebangs - should we consider accommodating for them in this tool? I'm curious as to whether or not any issues would arise out of inserting a copyright header before a shebang (whether it be issues from just our linter or otherwise) @jeremywiebe  thoughts? 🤔

Cdok:
Tests via `npm install` confirmed to be working on the SDK, noting that the glob needs to be wrapped in quotes to hit all of the nested directories.

Updated the readme and `--help` flag to reflect this

copyright/README.md
```diff
@@ -0,0 +1,45 @@
...
+Add the following to your `package.json` file
+
+```json
+"copyright:fix": "copyright './src/**/*.js' --fix",
+```
+
+Then run
+
+```bash
+npm run copyright:lint
```
jeremywiebe: This should be `copyright:fix` right?

copyright/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+This will add copyright headers to _all_ `.js` files in the `src` directory in the root of your project.
+
+## Lint Mode
+
+Passing the `--fix` flag to this tool will disable lint mode.
```
jeremywiebe: I think this is incorrect now also.  You pass nothing to lint and pass `--fix` to modify files.

jeremywiebe:
@Cdok I didn't think of shebangs... yes, we probably should account for them.

copyright/copyright.js
```diff
@@ -0,0 +1,128 @@
...
+                error = true
+            } else {
+                const contentStr = content.toString().split('\n')
+                let shebang = ''
+                let newData = ''
+
+                // accomodate for shebang and insert before header
+                if (contentStr[0].indexOf('#!') >= 0) {
+                    shebang = contentStr.shift()
+                    content = contentStr.join('\n')
```
Cdok: I'm not sure if this check for the shebang _feels_ right. It's currently working, but the string manipulation here feels a bit awkward.

screen shot 2017-05-04 at 4 15 12 pm

Cdok:
@jeremywiebe oneeeee more thing that I hadn't thought of 😏  

I just added an optional `--update` flag so that in 2018 the headers can be updated by passing the flag, or we could just use
```json
"copyright:lint": "copyright './src/**/*.js' --update",
```
in the `npm run` script, and let the updates happen automagically.

Either way, the option is there for both right now. 

Lastly, the shebang check functionality is up in this PR now too.

I've tested it against `bin` in the SDK.

With that said, I _think_ this PR is ready for the last feedback pass 👀

PR #162, opened 2017-09-08
Enforce ordering of imports

jeremywiebe:
Enforces that imports are grouped and separated with some logic. We use the default ordering as defined in this section of the docs: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#groups-array

`import/first` ensures that all imports occur before any statements.

Basically this:
```js
/* eslint import/order: ["error", {"newlines-between": "always"}] */
import fs from 'fs'
import path from 'path'

import sibling from './foo'

import index from './'
```

## Changes
- Adds enforcement of `import` ordering so that they have some structure and logic

## How To Test
- (necessary config changes)
- (necessary corresponding PRs)
- (how to access the new / changed functionality -- fixtures, URLs)

- (links, optional)

## TODOs:
- [ ] +1
- [ ] Updated README
- [ ] Updated CHANGELOG

### DIFF
--- a/es6/mobify-es6.yml
+++ b/es6/mobify-es6.yml
@@ -24,6 +24,12 @@ rules:
   import/no-unresolved: 'off'
   no-duplicate-imports: error
 
+  # Enforce some ordering of imports
+  import/order:
+    - error
+    - newlines-between: always
+  import/first: error
+
   # Stick with ES6 module syntax
   import/no-amd: error
   import/no-commonjs: error


fractaltheory:
Nitpick about your PR description - shouldn't the index import come after the sibling?

jeremywiebe:
@fractaltheory Ha! Yes.  I copied that directly from the docs for the plugin.

benlast:
The Mobify Python standard has this already: https://github.com/mobify/mobify-code-style/tree/develop/python#pep8

